PR #1, opened 2017-05-23
Add tables containing assembler directives and relocation expansions

michaeljclark:
Initial commit transferring across the tables from the rv8 repo.


michaeljclark:
Let me know if you think the examples from here are useful and I can make another commit.

- https://github.com/rv8-io/rv8/blob/master/doc/src/asm.md

palmer-dabbelt:
I think the examples would be good to have.

michaeljclark:
OK. I'll work on another commit...

PR #2, opened 2017-05-23
Add labels, adressing modes, LI, LA, constants and CSRs

michaeljclark:
This is the remainder from asm.md. The rest well have to work on adding...


palmer-dabbelt:
Thanks!

PR #7, opened 2018-03-02
Added basic table and general comments on registers

scotws:
Include table with most basic information on the registers of RV32I, with a basic skeleton for inclusion of floating point, control and vector registers. An alternative would be to only put RV32I here and include the other registers in separate chapters. Assumes that the use of `sp` and friends will be described in detail in later parts of the manual.

Signed-off-by: Scot W. Stevenson <scot.stevenson@gmail.com>

### DIFF
--- a/riscv-asm.md
+++ b/riscv-asm.md
@@ -19,7 +19,84 @@ duplicating it here.
 
 # Registers
 
-ISA and ABI register names for X, F, and CSRs.
+Registers are the most important part of any processor. RISC-V defines various
+types, depending on which extensions are included: The general registers (with
+the program counter), control registers, floating point registers (F extension),
+and vector registers (V extension).
+
+## General registers 
+
+The RV32I base integer ISA includes 32 registers, named `x0` to `x31`. The
+program counter `PC` is separate from these registers, in contrast to other
+processors such as the ARM-32. The first register, `x0`, has a special function:
+Reading it always returns 0 and writes to it are ignored. As we will see later,
+this allows various tricks and simplifications.
+
+In practice, the programmer doesn't use this notation for the registers. Though
+`x1` to `x31` are all equally general-use registers as far as the processor is
+concerned, by convention certain registers are used for special tasks. In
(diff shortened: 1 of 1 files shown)


riscv-asm.md
```diff
@@ -19,7 +19,85 @@ duplicating it here.
...
+program counter `PC` is separate from these registers, in contrast to other
+processors such as the ARM-32. The first register, `x0`, has a special function:
+Reading it always returns 0 and writes to it are ignored. As we will see later,
+this allows various tricks and simplifications.
+
+In practice, the programmer doesn't use this notation for the registers. Though
+`x1` to `x31` are all equally general-use registers as far as the processor is
+concerned, by convention certain registers are used for special tasks. The
+RISC-V **application binary interface** (ABI) standardizes this by giving the
+registers other "names" or "calling conventions". These are recognized by the
```
palmer-dabbelt: It's not really accurate to say the registers have calling conventions, it's the language that has a calling convention.

riscv-asm.md
```diff
@@ -19,7 +19,85 @@ duplicating it here.
...
+assembler and what you will usually see in code listings.
```
palmer-dabbelt: The `-M numeric` argument to objdump will provide numeric register names (ie, `x1`) instead of ABI register names.

riscv-asm.md
```diff
@@ -19,7 +19,85 @@ duplicating it here.
...
+_Registers of the RV32I. Based on RISC-V documentation and Patterson and
+Waterman "The RISC-V Reader" (2017)_
+
+As a general rule, the **saved registers** `s0` to `s11` are preserved across
+function calls, while the **argument registers** `a0` to `a7` and the
+**temporary registers** `t0` to `t6` are not.  The use of the various
+specialized registers such as `sp` by convention will be discussed later in more
+detail. 
+
+
```
palmer-dabbelt: Extra newlines

scotws:
Thanks! Included changes, though I'm still not sure if I got the part with the calling conventions right.

PR #15, opened 2018-11-08
Fix bug

missgya:
1. using register a0 to pass argument to function puts, not register a1.
2. "jalr ra, puts" would generate a compile error. Shoule be "jal puts" or "jal ra, puts".

### DIFF
--- a/riscv-asm.md
+++ b/riscv-asm.md
@@ -201,9 +201,9 @@ The following example shows how to load an absolute address:
 .section .text
 .globl _start
 _start:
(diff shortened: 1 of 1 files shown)


aswaterman:
Thanks - this looks correct to me. I’ll merge now, but would appreciate an ex post facto sanity check from @palmer-dabbelt

palmer-dabbelt:
Yep, the `jalr` was nonsense -- sorry that slipped through!  This should be `jal ra, puts`, not `jal puts`, as `jal puts` is an alias and we're trying to avoid those in the docs.

Thanks!

PR #16, opened 2018-11-08
Add new operator for R_RISCV_GOT_HI20

jrtc27:
Currently, the `la` pseudo-instruction cannot be expanded for position-independent code by the compiler, and it must emit the pseudo-instruction instead. This pull request introduces a new `%got_pcrel_hi` operator for this purpose.

### Name Bikeshedding

We have `%pcrel_hi` and `%tprel_hi` which would suggest `%got_hi` is a logical option. However, we also have `%tls_ie_pcrel_hi` and `%tls_gd_pcrel_hi` which would lead to `%got_pcrel_hi` being the new name. Despite `%got_hi` being possibly the more obvious choice at first, given that it's shorter, I am instead proposing `%got_pcrel_hi`. This is because `%tprel_hi` has its own `%tprel_lo`, whereas `%tls_ie_pcrel_hi` and `%tls_gd_pcrel_hi` reuse `%pcrel_lo`, much like this new operator, and therefore including `pcrel` in the name is both more consistent and makes it seem less odd that it's used in conjunction with `%pcrel_lo`. Thoughts?

### DIFF
--- a/riscv-asm.md
+++ b/riscv-asm.md
@@ -163,15 +163,19 @@ Directive    | Arguments                      | Description
 
 The following table lists assembler relocation expansions:
 
-Assembler Notation       | Description                 | Instruction / Macro
-:----------------------  | :---------------            | :-------------------
-%hi(symbol)              | Absolute (HI20)             | lui
-%lo(symbol)              | Absolute (LO12)             | load, store, add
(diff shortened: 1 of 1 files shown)


jim-wilson:
I'd suggest a patch to remove assembler operators from the psABI, since it duplicates what is here, and this info doesn't really belong in the psABI.

PR #33, opened 2020-06-13
Fix typo

Dahfizz9897:


aswaterman:
Not a typo.  `.globl` and `.global` do the same thing, with `.globl` being the more common form used by GCC.

Dahfizz9897:
Oh, interesting. I've never seen `.globl`

PR #37, opened 2020-06-24
riscv-asm.md: add missing pseudos for globals

ebahapo:


riscv-asm.md
```diff
@@ -377,6 +377,49 @@ for PIC as seen by objdump:
...
+Load and Store Global
+---------------------
+
+The following pseudo instructions are available to load from and store to
+global objects:
+
+  * `l{b|h|w|d} <rd>, <symbol>`: load byte, half word, word or double word from global
+  * `s{b|h|w|d} <rd>, <symbol>, <rt>`: store byte, half word, word or double word to global
+  * `fl{w|d} <rd>, <symbol>, <rt>`: load float or double from global
+  * `fs{w|d} <rd>, <symbol>, <rt>`: store float or double to global
```
jim-wilson, line 389: This also works for flq/fsq for the Q extension, though I don't know of anyone using that yet.

riscv-asm.md
```diff
@@ -377,6 +377,49 @@ for PIC as seen by objdump:
...
 ```
 
+Load and Store Global
+---------------------
+
+The following pseudo instructions are available to load from and store to
+global objects:
+
+  * `l{b|h|w|d} <rd>, <symbol>`: load byte, half word, word or double word from global
+  * `s{b|h|w|d} <rd>, <symbol>, <rt>`: store byte, half word, word or double word to global
```
jim-wilson, line 387: We don't have support for lq/sq for rv128i yet, but if we did, I would expect them to work the same way.

riscv-asm.md
jim-wilson: There are flh/fsh instructions for the Zfh half-word float format, but this is still an unofficial draft proposal.

riscv-asm.md
```diff
@@ -377,6 +377,49 @@ for PIC as seen by objdump:
...
+
+```assembly
+	lw	a0, var1
+	fld	fa0, var2, t0
+	sw	a0, var3, t0
+	fsd	fa0, var4, t0
+```
+
+Which generates the following assembler output and relocations
+as seen by `objdump`:
```
jim-wilson, line 401: This is specifically medany output which should perhaps be mentioned, as you will see different code with other code models.

ebahapo: Again, are you sure that it's not the same in `medlow`?

jim-wilson: I checked gas, and see that you are right.  Though I'm tempted to call this a bug.  We normally shouldn't be using auipc in medlow code, except for calls where we always use auipc..  Though as I mentioned elsewhere, medany code can be linked into medlow code, and the result is medlow, so this is safe.  It is just unexpected.

riscv-asm.md
```diff
@@ -377,6 +377,49 @@ for PIC as seen by objdump:
...
 
+Load and Store Global
+---------------------
+
+The following pseudo instructions are available to load from and store to
+global objects:
+
+  * `l{b|h|w|d} <rd>, <symbol>`: load byte, half word, word or double word from global
+  * `s{b|h|w|d} <rd>, <symbol>, <rt>`: store byte, half word, word or double word to global
+  * `fl{h|w|d|q} <rd>, <symbol>, <rt>`: load half, float, double or quad precision from global
```
lenary, line 388: Should the loads here have 3 arguments?

jim-wilson: For a load, rd is used as the temp register.  But that doesn't work for a store, so the temp register has to be specified explicitly.  And for a float load, we can't use an fp reg as an address temp, so it needs an explicit int temp reg also.

lenary: Ah, so actually all the three instruction variants, including `s{b|h|w|d}`, use the third instruction as a temporary. Given I missed this, it might be useful to add a sentence just clarifying this (though I now see you can work it out from the code example shown).

riscv-asm.md
ebahapo: How do you specify the code model to `gas`?  Did I miss it in its man page?

jim-wilson: You can't specify the code model to gas.  Looks like an oversight, which is why the macros all emit medany code, because medany code works for both medlow and medany.  But that won't work for compact or FDPIC, which means those macros won't be usable with non-medlow/medany code models unless we change how this works.

jim-wilson: Or maybe won't work with other code models, depending on the amount of compatibility with the existing medlow/medany code models.

jim-wilson:
Merging this in the interest of forward progress.

There is a comment from Sam Elliot about clarifying why int load doesn't need a temp reg.

There is a comment from me about how all asm macros emitting medany code might be a bug, particularly when we add a third code model.

I think these can be addressed as follow on issues or pull requests.

PR #39, opened 2020-06-27
riscv-asm.md: simplify examples

ebahapo:


riscv-asm.md
```diff
@@ -320,21 +295,20 @@ Load Immediate
...
 
-```
-   0:   deadc537                lui     a0,0xdeadc
-   4:   eef50513                addi    a0,a0,-273
+```assembly
+0000000000000000 <.text>:
+   0:	00038537          	lui	a0,0x38
+   4:	ab75051b          	addiw	a0,a0,-1353
+   8:	00e51513          	slli	a0,a0,0xe
+   c:	eef50513          	addi	a0,a0,-273 # 37eef <CONSTANT-0xdeaa4000>
```
jim-wilson, line 311: This is the rv64 sequence for a non-sign extended 32-bit constant (without the B extension).  The comment above says rv32i.  For rv64i, the constant should be 0xffffffffdeadbeef and you get the same two instructions as rv32i.

riscv-asm.md
```diff
@@ -229,89 +229,64 @@ Absolute addressing
...
-	    addi a0, a0,  %lo(msg)       # load msg(lo)
-	    jal ra, puts
-2:	    j 2b
-
-.section .rodata
-msg:
-	    .string "Hello World\n"
+```assembly
+	lui	a0, %hi(msg + 1)
+	addi	a0, a0, %lo(msg + 1)
```
jim-wilson, line 234: The msg+1 is a little confusing.  It makes sense if you are trying to document how offsets are handled.  But the text above says "an absolute address" which might make someone wonder where the +1 is coming from.  Perhaps that could be "the absolute address msg+1" to clarify where the +1 comes from.  And similarly with the examples below.

ebahapo: It's just that `addi a0, a0, 0` is displayed as `mv a0, a0`, which is not intuitive to have a relocation attached to it.  Then again, since RISC-V uses RELA, why is the addend encoded in the instruction?

jim-wilson: You can avoid the mv by using -Mno-aliases with objdump.  That disables all instruction aliases, which can be occasionally confusing, but for these simple examples that will probably be OK.

Good question about the addend in the instruction.  There is code to handle relocs with constant values, write them into the instruction and mark them done if no symbol address in the reloc.  But it probably should check for the symbol address first, and only write into the insn if no symbol address is in the reloc.  Looks like a minor harmless oversight.  I tried to trace it back, but it is there in 2015 which is before I started working on it and before it was upstream, so it will be hard to figure out happened.  The old history isn't well preserved.

ebahapo: Using `-Mno-alias` with `objdump` didn't affect `mv a0, a0`.

lenary: I *think* the command line is `-M no-aliases` - the plural catches me out all the time too.

jim-wilson:
```
rohan:2038$ riscv32-unknown-elf-objdump -dr a.out

a.out:     file format elf32-littleriscv


Disassembly of section .text:

00000000 <.text>:
   0:	00000537          	lui	a0,0x0
			0: R_RISCV_HI20	msg
			0: R_RISCV_RELAX	*ABS*
   4:	00050513          	mv	a0,a0
			4: R_RISCV_LO12_I	msg
			4: R_RISCV_RELAX	*ABS*
rohan:2039$ riscv32-unknown-elf-objdump -dr -Mno-aliases a.out

a.out:     file format elf32-littleriscv


Disassembly of section .text:

00000000 <.text>:
   0:	00000537          	lui	a0,0x0
			0: R_RISCV_HI20	msg
			0: R_RISCV_RELAX	*ABS*
   4:	00050513          	addi	a0,a0,0 # 0x0
			4: R_RISCV_LO12_I	msg
			4: R_RISCV_RELAX	*ABS*
rohan:2040$ 
```

PR #41, opened 2020-07-03
Add new definitions for the compact code model

ebahapo:


riscv-asm.md
```diff
@@ -202,8 +202,16 @@ Assembler Notation          | Description                    | Instruction / Mac
 %tls_ie_pcrel_hi(symbol) \* | TLS IE "Initial Exec" (HI20)   | auipc
 %tls_gd_pcrel_hi(symbol) \* | TLS GD "Global Dynamic" (HI20) | auipc
 %got_pcrel_hi(symbol) \*    | GOT PC-relative (HI20)         | auipc
+%gprel_hi(symbol)           | GP-relative (HI20)             | lui
+%gprel_lo(symbol)           | GP-relative (LO12)             | load, store, add
```
lenary, line 206: What is the reason to skip `%gprel_add(symbol)` (and `%got_gprel_add(symbol)`)?

ebahapo: That's what `%gprel(symbol)` and `%got_gprel(symbol)` are for as well.  They are overloaded for `add`, `la`, loads, stores.

lenary: What if I want to write out those sequences longhand, like you can with the tprel sequences? Something like the following in the .S file:

```
  lui a0, %gprel_hi(symbol+off)
  add a0, a0, gp, %gprel_add(symbol+off)
  addi a0, a0, %gprel_lo(symbol+off)
```

Is this something we want to allow?

ebahapo: Most definitely.

```
lui a0, %gprel_hi(symbol+off)
add a0, a0, gp, %gprel(symbol+off)
addi a0, a0, %gprel_lo(symbol+off)
```

lenary: Ok, neat. My preference is for the former, because that mostly matches what has come before with the other relocation types.

ebahapo: For the sake of uniformity, would you rather prefer `add a0, a0, gp, %gprel_add(symbol+off)` over `add a0, a0, gp, %gprel(symbol+off)`?

lenary: Potentially, but maybe we can parse both anyway, and have `%gprel(symbol+off)` as canonical?

jrtc27: Having `%gprel` be canonical is sad. I think it would be much better to have `%gprel(...)` be purely used in the pseudoinstructions, and maintain consistency otherwise by using `%gprel_add` to tag the `add` instruction. Overloading meaning adds needless mental overhead, whereas some semblance of consistency makes it much easier to move between models.

I personally don't like using `la rd, %gprel(symbol)` etc even, and would rather those be separate pseudoinstructions entirely. But that's a different discussion and more personal preference than actual concern.

riscv-asm.md
```diff
@@ -377,6 +385,50 @@ for PIC as seen by objdump:
...
+as seen by `objdump`:
+
+```assembly
+0000000000000000 <.text>:
+   0:	00000537          	lui	a0,0x0
+			0: R_RISCV_GOT_GPREL_HI20	msg+0x1
+   4:	00350533          	add	a0,a0,gp
+			4: R_RISCV_GOT_GPREL_ADD	msg+0x1
+   8:	00153503          	ld	a0,1(a0) # 0x1
+			8: R_RISCV_GOT_GOT_GPREL_LO12_I	msg+0x1
```
jrtc27, line 429: Duplicate GOT_

riscv-asm.md
```diff
@@ -377,6 +385,50 @@ for PIC as seen by objdump:
 			4: R_RISCV_PCREL_LO12_I	.L0
 ```
 
+When used with the `%gprel` relocation function,
+the `la` pseudo instruction
+is used to load symbol addresses relative to the `gp`,
+as in the compact code model:
+
+```assembly
+	la a0, %gprel(msg + 1)
```
jrtc27, line 394: I don't like that `la` depends on -fPIC except suddenly when you throw `%gprel` at it. I'd expect this to use the GOT for -fPIC and GP directly otherwise. I feel you really want `lla` here; this model _can't_ work with preemptible symbols, but `la` does, so this is dangerous.

riscv-asm.md
```diff
@@ -377,6 +385,50 @@ for PIC as seen by objdump:
...
+			8: R_RISCV_GPREL_LO12_I	msg+0x1
+```
+
+When used with the `%got_gprel` relocation function,
+the `la` pseudo instruction
+is used to load symbol addresses in the GOT relative to the `gp`,
+as in the compact code model:
+
+```assembly
+	la a0, %got_gprel(msg + 1)
```
jrtc27, line 416: This is also surprising. It's no longer -fPIC dependent.

riscv-asm.md
ebahapo: I don't have a problem with `%gprel_add`.

riscv-asm.md
jrtc27: Why was this marked as resolved? The issue still stands.

ebahapo: Please, stand by as I upload the new patch.

ebahapo:
Ping, please.

jrtc27:
I don't think this should be merged until it's been approved by a working group? It's not just clarifying what the current Unix ABI does.

riscv-asm.md
```diff
@@ -384,6 +494,97 @@ as seen by `objdump`:
...
+			8: R_RISCV_TLS_GOT_GPREL_LO12	msg+0x1
+```
+
+The followig pseudo instruction is used for the TLS global dynamic,
+relative to `\_\_global_pointer$`,
+assumed to be specified in the last register
+or defaulting to the `gp` register:
+
+```assembly
+	la.tls.gd.gprel	a0, msg + 1, s0
```
jrtc27, line 572: Why does this one get a new mnemonic when the others use `%(got_)gprel` on the expression?

ebahapo: Though there are the relocation functions `%tls_{gd,ie}_pcrel_hi`, there are also the pseudo instructions `la.tls.{gd,ie}`, which are implictiy PC relative.  This pseudo instruction is equivalent to the latter, but explicitly relative to the `__global_pointer$`.

ebahapo:
Ping, pretty please.

PR #46, opened 2020-08-04
Remove old function call section

y2kbugger:
Looks like it was left over, maybe from a previous merge?


aswaterman:
Naturally, they aren't exact duplicates, and one of them doesn't strictly dominate the other :-\ I'll manually merge them.

aswaterman:
Whoops, I misread the diff.  The one you removed does dominate the other.

PR #55, opened 2020-09-23
Add a link to a SO discussion of callee/caller saved registers.

gevorgyana:
As a newcomer, I found it difficult to grasp at first.


jrtc27:
Links to Q&A sites aren't appropriate for technical documentation, and such terminology is easy to look up.

PR #56, opened 2020-09-23
Add a table summarizing the available pseudoinstructions.

### LINKED ISSUE #54: Cover more pseudoinstructions, as it was done in riscv-spec-v2.2
This issue https://github.com/riscv/riscv-isa-manual/issues/540 suggested to supersede the section on pseudoinstructions inside RISC-V ISA manual with riscv-asm-manual. We can cover more pseudoinstructions than we do now in this repository - I can't see any mention of `fence`, no mention of "`branch`" pseudoinstructions. Shouldn't we just add the table from the ISA manual 2.2 to the markdown?

gevorgyana:
#fixes #54

### DIFF
--- a/riscv-asm.md
+++ b/riscv-asm.md
@@ -24,7 +24,7 @@ types, depending on which extensions are included: The general registers (with
 the program counter), control registers, floating point registers (F extension),
 and vector registers (V extension).
 
-## General registers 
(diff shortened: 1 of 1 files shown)


aswaterman:
Looks good to me!

PR #59, opened 2020-10-09
Add PAUSE hint

aswaterman:
Merge whenever https://github.com/riscv/riscv-isa-manual/pull/398 is merged.


kito-cheng:
Oooh, this hold here so long time!?

PR #60, opened 2020-10-18
Remove broken link

mehmetoguzderin:
Seems like privileged spec URL is no longer working and is listed in the same page with User Level specification


aswaterman:
Thanks.

PR #61, opened 2020-12-03
Add pseudoinstruction: sext.b, sext.h, zext.b, zext.h and zext.w

kito-cheng:
Add `sext.b`, `sext.h`, `zext.b`, `zext.h` and `zext.w`, and mention `sext.b` and `sext.h` could be non pseudo instruction, the motivation is we already has `sext.w` now, and B ext also want to define more pseudo for `sext` or `zext`, so why not add to the baseline pseudo instruction, and let assembler pick best way to expand.

### DIFF
--- a/riscv-asm.md
+++ b/riscv-asm.md
@@ -532,8 +532,8 @@ fail_msg:
 
 ## A listing of standard RISC-V pseudoinstructions
 
-Pseudoinstruction            | Base Instruction(s)                                           | Meaning
-:----------------------------|:--------------------------------------------------------------|:-------
+Pseudoinstruction            | Base Instruction(s)                                           | Meaning   | Comment
+:----------------------------|:--------------------------------------------------------------|:----------|:--------|
 la rd, symbol                | auipc rd, symbol[31:12]; addi rd, rd, symbol[11:0]            | Load address
 l{b\|h\|w\|d} rd, symbol     | auipc rd, symbol[31:12]; l{b\|h\|w\|d} rd, symbol[11:0]\(rd\) | Load global
 s{b\|h\|w\|d} rd, symbol, rt | auipc rt, symbol[31:12]; s{b\|h\|w\|d} rd, symbol[11:0]\(rt\) | Store global
@@ -545,7 +545,12 @@ mv rd, rs                    | addi rd, rs, 0                                   
 not rd, rs                   | xori rd, rs, -1                                               | One’s complement
 neg rd, rs                   | sub rd, x0, rs                                                | Two’s complement
 negw rd, rs                  | subw rd, x0, rs                                               | Two’s complement word
+sext.b rd, rs                | slli rd, rs, XLEN - 8; srai rd, rd, XLEN - 8                  | Sign extend byte | It will expand another instruction sequence when B extension is available*[1]
+sext.h rd, rs                | slli rd, rs, XLEN - 16; srai rd, rd, XLEN - 16                | Sign extend half word | It will expand another instruction sequence when B extension is available*[1]
 sext.w rd, rs                | addiw rd, rs, 0                                               | Sign extend word
+zext.b rd, rs                | andi rd, rs, 255                                              | Zero extend byte
+zext.h rd, rs                | slli rd, rs, XLEN - 16; srli rd, rd, XLEN - 16                | Zero extend half word | It will expand another instruction sequence when B extension is available*[1]
+zext.w rd, rs                | slli rd, rs, XLEN - 32; srli rd, rd, XLEN - 32                | Zero extend word | It will expand another instruction sequence when B extension is available*[1]
 seqz rd, rs                  | sltiu rd, rs, 1                                               | Set if = zero
 snez rd, rs                  | sltu rd, x0, rs                                               | Set if != zero
 sltz rd, rs                  | slt rd, rs, x0                                                | Set if < zero
@@ -574,6 +579,8 @@ ret                          | jalr x0, x1, 0                                   
 call offset                  | auipc x6, offset[31:12]; jalr x1, x6, offset[11:0]            | Call far-away subroutine
 tail offset                  | auipc x6, offset[31:12]; jalr x0, x6, offset[11:0]            | Tail call far-away subroutine
 fence                        | fence iorw, iorw                                              | Fence on all memory and I/O
+
+* [1] We don't specific code sequence for B-extension, since B-extension still not ratified or frozen, we'll specify the expansion sequence once it's frozen.
 
 ## Pseudoinstructions for accessing control and status registers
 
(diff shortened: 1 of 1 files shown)


kito-cheng:
@clairexen @aswaterman

aswaterman:
Seems reasonable to me, to simplify some assembly programming.  However, compilers should avoid avoid sext.[b|h] and zext.[h|w] when Zbb isn't present (though actually Zba can implement zext.w, too), since it can be more efficient to schedule the two shifts apart from each other.

kito-cheng:
> However, compilers should avoid avoid sext.[b|h] and zext.[h|w] when Zbb isn't present (though actually Zba can implement zext.w, too), since it can be more efficient to schedule the two shifts apart from each other.

Agree with that, generally compiler should avoid emit pseudo instruction, which will expand to more than one instructions...unless necessary like `call` or `tail`.

jim-wilson:
Zba has zext.h also, so all of them are a single instruction with the B extension.

This creates work for binutils and llvm developers, and maybe FreeBSD also.  We can handle the binutils work.  Who does the llvm work?  It would be nice to get an OK from the LLVM folks at least before committing this.

kito-cheng:
@lenary @ebahapo 

> We can handle the binutils work. Who does the llvm work?

We have few LLVM developer now, and it should not too much effort I guess, maybe @topperc could help on that?

kito-cheng:
Note: I didn't mention how pseudo expand with B-extension on purpose, since it still might change.

topperc:
I agree it shouldn't be much effort in LLVM. I can help.

kito-cheng:
@topperc thanks you take this :)

I guess just wait some another LLVM folks (maybe non-SiFive guys?) to ack this.

lenary:
I don't have a fundamental problem with this, and would be happy to help review a patch :)

lenary:
I do wonder if we want to say "at most two 32-bit instructions" or the like, in these docs? It's unclear to me what guarantees people coding with pseudo-instructions want - turning off relaxation and compressed instructions is needed even if you stick to non-pseduo-instructions and want predictable instruction sizes - presumably if you require predictable instruction sizes, it's also best to stay away from pseudo-instructions too.

knightsifive:
From my perspective as an assembly programmer who uses pseudoinstructions, all I really want are clear descriptions of

- the different code sequences that the assembler may emit (vs. saying "myriad sequences"), and 
- the different linker relaxations that may be subsequently applied (vs. saying "go read the `ld` source tree").

In terms of "guarantees", what I'm looking for is a guarantee that "(only) the following things may happen" to my assembly program. What isn't clear to me is whether these things should be considered implementation details --- thus belonging in assembler/linker documentation --- or part of the assembly language specification --- thus belonging here. 

I completely understand that we are woefully understaffed w.r.t. documentation, I'm not asking anyone to do anything.

lenary:
@kito-cheng fyi, I revise my offer to "review" a patch that implements this, I've got an LLVM patch mostly ready which I'll upload to phabricator and tag you as a reviewer for.

kito-cheng:
Changes:
- Add more description about we might emit another code sequence when B extension is enabled, but not specify the code sequence at this moment, because it's not frozen yet.

kito-cheng:
@lenary  I saw the patch, thanks!

riscv-asm.md
```diff
@@ -545,7 +545,12 @@ mv rd, rs                    | addi rd, rs, 0
 not rd, rs                   | xori rd, rs, -1                                               | One’s complement
 neg rd, rs                   | sub rd, x0, rs                                                | Two’s complement
 negw rd, rs                  | subw rd, x0, rs                                               | Two’s complement word
+sext.b rd, rs                | slli rd, rs, XLEN - 8; srai rd, rd, XLEN - 8                  | Sign extend byte | It will expand another instruction sequence when B extension is available*[1]
```
lenary, line 548:
```suggestion
sext.b rd, rs                | slli rd, rs, XLEN - 8; srai rd, rd, XLEN - 8                  | Sign extend byte | It will expand to another instruction sequence when B extension is available*[1]
```

riscv-asm.md
```diff
@@ -545,7 +545,12 @@ mv rd, rs                    | addi rd, rs, 0
...
+sext.h rd, rs                | slli rd, rs, XLEN - 16; srai rd, rd, XLEN - 16                | Sign extend half word | It will expand another instruction sequence when B extension is available*[1]
```
lenary, line 549:
```suggestion
sext.h rd, rs                | slli rd, rs, XLEN - 16; srai rd, rd, XLEN - 16                | Sign extend half word | It will expand to another instruction sequence when B extension is available*[1]
```

riscv-asm.md
```diff
@@ -545,7 +545,12 @@ mv rd, rs                    | addi rd, rs, 0
...
 sext.w rd, rs                | addiw rd, rs, 0                                               | Sign extend word
+zext.b rd, rs                | andi rd, rs, 255                                              | Zero extend byte
+zext.h rd, rs                | slli rd, rs, XLEN - 16; srli rd, rd, XLEN - 16                | Zero extend half word | It will expand another instruction sequence when B extension is available*[1]
```
lenary, line 552:
```suggestion
zext.h rd, rs                | slli rd, rs, XLEN - 16; srli rd, rd, XLEN - 16                | Zero extend half word | It will expand to another instruction sequence when B extension is available*[1]
```

riscv-asm.md
```diff
@@ -545,7 +545,12 @@ mv rd, rs                    | addi rd, rs, 0
...
+zext.w rd, rs                | slli rd, rs, XLEN - 32; srli rd, rd, XLEN - 32                | Zero extend word | It will expand another instruction sequence when B extension is available*[1]
```
lenary, line 553:
```suggestion
zext.w rd, rs                | slli rd, rs, XLEN - 32; srli rd, rd, XLEN - 32                | Zero extend word | It will expand to another instruction sequence when B extension is available*[1]
```

riscv-asm.md
```diff
@@ -575,6 +580,8 @@ call offset                  | auipc x6, offset[31:12]; jalr x1, x6, offset[11:0
 tail offset                  | auipc x6, offset[31:12]; jalr x0, x6, offset[11:0]            | Tail call far-away subroutine
 fence                        | fence iorw, iorw                                              | Fence on all memory and I/O
 
+* [1] We don't specific code sequence for B-extension, since B-extension still not ratified or frozen, we'll specify the expansion sequence once it's frozen.
```
lenary, line 583:
```suggestion
* [1] We don't specify the code sequence when the B-extension is present, since B-extension still not ratified or frozen. We will specify the expansion sequence once it's frozen.
```

lenary (approved):
Some grammar nits, but LGTM.

kito-cheng:
@lenary Changes applied, thanks!

lenary:
I just landed the LLVM implementation in https://github.com/llvm/llvm-project/commit/12406ade0625

PR #64, opened 2021-05-12
Add a new directive .variant_cc for variant calling conventions.

Hsiangkai:


ebahapo:
Instead of a new directive, perhaps a new `.type` argument would make more sense, since functions typically use `.type` to specify function symbols.

Hsiangkai:
> Instead of a new directive, perhaps a new `.type` argument would make more sense, since functions typically take a `.type` to be specified as function symbols.

ARM defines a new directive, .variant_pcs, for the purpose. Should we extend the syntax of .type?

* https://sourceware.org/binutils/docs/as/AArch64-Directives.html
* https://reviews.llvm.org/D89138

kito-cheng:
@jim-wilson @jrtc27

kito-cheng (approved):
LGTM.

jim-wilson (approved):
We shouldn't invent anything new if we don't have to.  .variant_cc is consistent with the aarch64 port.  This looks good to me.

kito-cheng:
@asb @jrtc27 @luismarques @MaskRay 

This PR is related to `Calling conventions for the lazily bound functions.` https://github.com/riscv/riscv-elf-psabi-doc/pull/190, used to mark a symbol is `STO_RISCV_VARIANT_CC`, and o not lazy binding for this symbol.

I guess there is no much controversy here since the solution just same as AArch64, but I would like to got some LGTM from LLVM community.

asb:
> @asb @jrtc27 @luismarques @MaskRay
> 
> This PR is related to `Calling conventions for the lazily bound functions.` [riscv/riscv-elf-psabi-doc#190](https://github.com/riscv/riscv-elf-psabi-doc/pull/190), used to mark a symbol is `STO_RISCV_VARIANT_CC`, and o not lazy binding for this symbol.
> 
> I guess there is no much controversy here since the solution just same as AArch64, but I would like to got some LGTM from LLVM community.

I've got no objections, but I'd defer to @maskray on this one.

MaskRay:
LGTM (I don't have github permission to approve).

Using a new directive (not reusing `.type`) has the advantages:

* `.type` is for st_type. st_other is orthogonal to st_type.
* avoid overriding the generic `.type` with arch-specific semantics.

I assume that we cannot reuse `.variant_pcs` because PCS is an aarch64 term...
A new directive theoretically allows extension without introducing user confusion, though it may be hard to think how `.variant_cc` could be extended in the future.

kito-cheng:
@jim-wilson @aswaterman could you merge this PR? I believe we reached a consensus between LLVM and GNU toolchain folks, and the psabi spec merged too, https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/190

jrtc27:
Commits should be squashed and reworded to not talk about bind now

Hsiangkai:
Commits are squashed and the title is changed. Sorry for the late update.

PR #67, opened 2021-07-09
Proposal:  Extend .option directive for control enabled extensions on specific code region

kito-cheng:
Changes:
- 2021/8/12:
  - Add , after arch.
  - Only allow enable/disable one arch one time.
  - Add = operator to set arch to specific configuration.
- 2021/7/29: Extend `.option` rather than `.push_arch`/`.pop_arch`.

------
This commit extend .option diretive with new keyword:

- .option arch <ext-list>

Enable and/or disable specific ISA extensions for following code regions, but 
without changing the arch attribute, that means it won't raise the minimal
execution environment requirement, so the user should take care to protect
execution of the code regions around `.option push`/`.option arch`/`.option pop`.
A typical use case is is with `ifunc`, e.g. the libc is built with `rv64gc`,
but a few functions like memcpy provide two versions, one built with `rv64gc`
and one built with `rv64gcv`, and then select between them by ifunc mechanism
at run-time.  However we don't want to change the minimal execution environment
requirement to `rv64gcv`, since the `rv64gcv` version will be invoked only if
the execution environment supports the vector extension, so the minimal
execution environment requirement still is `rv64gc`.

Example:
```assembly
.attribute arch, "rv64imafdc"
 # You can only use instruction from i, m, a, f, d and c extensions.
memcpy_general:
    add     a5,a1,a2
    beq     a1,a5,.L2
    add     a2,a0,a2
    mv      a5,a0
.L3:
    addi    a1,a1,1
    addi    a5,a5,1
    lbu     a4,-1(a1)
    sb      a4,-1(a5)
    bne     a5,a2,.L3
.L2:
    ret 

.option push    # Push current option to stack.
.option arch +v # Enable vector extension, we can use any instruction in imafdcv extension.
memcpy_vec:
    mv a3, a0
.Lloop:
    vsetvli t0, a2, e8, m8, ta, ma
    vle8.v v0, (a1)
    add a1, a1, t0
    sub a2, a2, t0
    vse8.v v0, (a3)
    add a3, a3, t0
    bnez a2, .Lloop
    ret
.option pop   # Pop current option from stack, restore the enabled ISA extension status to imafdc.
```

----
Notes:
- This proposal are expected work with [mapping symbol](https://github.com/riscv/riscv-elf-psabi-doc/pull/196), which is also in proposal stage.
- This PR are depend on https://github.com/riscv/riscv-asm-manual/pull/68, and will rebase after that merge.

### DIFF
--- a/riscv-asm.md
+++ b/riscv-asm.md
@@ -164,7 +164,7 @@ Directive    | Arguments                      | Description
 .macro       | name arg1 [, argn]             | begin macro definition \argname to substitute
 .endm        |                                | end macro definition
 .type        | symbol, @function              | accepted for source compatibility
-.option      | {rvc,norvc,pic,nopic,relax,norelax,push,pop} | RISC-V options. Refer to [.option](#.option) for a more detailed description.
+.option      | {arch,rvc,norvc,pic,nopic,relax,norelax,push,pop} | RISC-V options. Refer to [.option](#.option) for a more detailed description.
 .byte        | expression [, expression]*     | 8-bit comma separated words
 .2byte       | expression [, expression]*     | 16-bit comma separated words
 .half        | expression [, expression]*     | 16-bit comma separated words
@@ -214,7 +214,154 @@ ATTRIBUTE_VALUE := <string>
 
 #### `rvc`/`norvc`
 
-Enable/disable the C-extension for the following code region.
+This option will be deprecated soon after `.option arch` has been widely
+implemented on main stream open source toolchains.
+
+Enable/disable the C-extension for the following code region. This option is
+equivalent to `.option arch, +c`/`.option arch, -c`, but widely support by
+older toolchain versions.
+
+Alternative style:
+
+```
+.option push
+.option arch, +c   # Alternative of .option rvc
+.option pop
+
+.option push
+.option arch, -c   # Alternative of .option norvc
+.option pop
+```
+
+NOTE: `.option rvc` might set the ELF flag `EF_RISCV_RVC` in some toolchains. That
+might cause the linker to compress instructions in code regions where that was
+not intended.
+
+NOTE: There is a difference between `.option rvc`/`.option norvc` and
+`.option arch, +c`/`.option arch, -c`. The latter won't set EF_RISCV_RVC in the
+ELF flags.
+
+#### `arch`
+
+Enable and/or disable specific ISA extensions for the following code regions, but
+without changing the arch attribute and `EF_RISCV_RVC` in the ELF flags, that
+means it will not raise the minimal execution environment requirement, so the user
+should take care to the execution of the code regions around
+`.option push`/`.option arch`/`.option pop`.
+
+Syntax for `.option arch`:
+```
+.option arch, <EXTENSIONS-OR-FULLARCH>
+
+EXTENSIONS-OR-FULLARCH := <EXTENSIONS>
+                        | <FULLARCHSTR>
+
+EXTENSIONS             := <EXTENSION> ',' <EXTENSIONS>
+                        | <EXTENSION>
+
+FULLARCHSTR            := <full-arch-string>
+
+EXTENSION              := <OP> <EXTENSION-NAME> <VERSION>
+
+OP                     := '+'
+                        | '-'
+
+VERSION                := [0-9]+ 'p' [0-9]+
+                        | [1-9][0-9]*
+                        |
+
+EXTENSION-NAME         := Naming rule is defined in RISC-V ISA manual
+```
+
+- Extension version can be omitted, the assembler will use the built-in default
+  version for that extension.
+- `OP` can be enable (`+`) or disable (`-`).
+- Format of `<full-arch-string>` is the same as `-march` option.
+
+Example:
+
+```assembly
+.attribute arch, rv64imafdc
+# You can only use instructions from the i, m, a, f, d and c extensions.
+memcpy_general:
+    add     a5,a1,a2
+    beq     a1,a5,.L2
+    add     a2,a0,a2
+    mv      a5,a0
+.L3:
+    addi    a1,a1,1
+    addi    a5,a5,1
+    lbu     a4,-1(a1)
+    sb      a4,-1(a5)
+    bne     a5,a2,.L3
+.L2:
+    ret
+
+.option push     # Push current options to the stack.
+.option arch, +v # Enable vector extension, we can use any instruction in imafdcv extension.
+memcpy_vec:
+    mv a3, a0
+.Lloop:
+    vsetvli t0, a2, e8, m8, ta, ma
+    vle8.v v0, (a1)
+    add a1, a1, t0
+    sub a2, a2, t0
+    vse8.v v0, (a3)
+    add a3, a3, t0
+    bnez a2, .Lloop
+    ret
+.option pop   # Pop current option from the stack, restore the enabled ISA extension status to imafdc.
+
+.option push     # Push current option to the stack.
+.option arch, -c # Disable compressed extension, we can't use any instruction in extension.
+memcpy_norvc:
+    add     a5,a1,a2
+    beq     a1,a5,.L2
+    add     a2,a0,a2
+    mv      a5,a0
+.L3:
+    addi    a1,a1,1
+    addi    a5,a5,1
+    lbu     a4,-1(a1)
+    sb      a4,-1(a5)
+    bne     a5,a2,.L3
+.L2:
+    ret
+.option pop   # Pop current option from the stack, restore the enabled ISA extension status to imafdc.
+
+.option push  # Push current option to the stack.
+.option arch, rv64imc # Set arch to rv64imc.
+    nop
+.option pop   # Pop current option from the stack, restore the enabled ISA extension status to imafdc.
+
+```
+
+NOTE: A typical use case is with `ifunc`, e.g. the C library is built with
+`rv64gc`, but a few functions like memcpy provide two versions, one built with
+`rv64gc` and one built with `rv64gcv`, and then select between them by ifunc
+mechanism at run-time.  However, we don't want to change the minimal execution
+environment requirement to `rv64gcv`, since the `rv64gcv` version will be
+invoked only if the execution environment supports the vector extension, so
+the minimal execution environment requirement still is `rv64gc`.
+
+NOTE: `.option arch, +` will also enable all required extensions, for example,
+`rv32i` + `.option arch, +v` will also enable `f`, `d`, `zve32x`, `zve32f`,
+`zve64x`, `zve64f`, `zve64d`, `zvl32b`, `zvl64b` and `zvl128b` extensions.
+
+NOTE: We recommend `.option arch, +` and `.option arch, -` are used with
+`.option push`/`.option pop` instead of a `.option arch, +` / `.option arch, -`
+pair, because `.option arch, +` will enable all required extensions, but
+`.option arch, -` only disables the specific extension, so the result might be
+unexpected, for example: `rv32i` + `.option arch, +v` + `.option arch, -v`
+will result `rv32ifd_zve32x_zve32f_zve64x_zve64f_zve64d_zvl32b_zvl64b_zvl128b`
+not `rv32i`.
+Another example is `.option arch, rv64ifd` + `.option arch, -f` will still got
+same result `rv64ifd`, because `f` will add back during compute implication
+extensions for `rv64id`.
+
+NOTE: `.option arch, +<ext>, -<ext>` is accepted and will result in enabling the
+extensions that depend on `ext`, e.g. `rv32i` + `.option arch, +v, -v` will result
+`rv32ifd_zve32x_zve32f_zve64x_zve64f_zve64d_zvl32b_zvl64b_zvl128b`.
 
 #### `pic`/`nopic`
 
@@ -226,10 +373,10 @@ affect the expansion of the `la` pseudoinstruction, refer to
 
 Enable/disable linker relaxation for the following code region.
 
-NOTE: Code region follows by `.option relax` will emit
-`R_RISCV_RELAX`/`R_RISCV_ALIGN` even linker unsupport relaxation, suggested
-usage is using `.option norelax` with `.option push`/`.option pop` if
-you want to disable linker relaxation on specific code region.
+NOTE: A code region followed by `.option relax` will emit
+`R_RISCV_RELAX`/`R_RISCV_ALIGN` even if the linker does not support relaxation.
+The suggested usage is using `.option norelax` with `.option push`/`.option pop`
+if linker relaxation should be disabled for a code region.
 
 NOTE: Recommended way to disable linker relaxation of specific code region is
 use `.option push`, `.option norelax` and `.option pop`, that prevent enabled


riscv-asm.md
```diff
@@ -182,6 +182,67 @@ Directive    | Arguments                      | Description
...
+### <a name=push-pop-arch></a>`.push_arch`/`.pop_arch`
+
+`.push_arch` / `.pop_arch` are used to control enabled ISA extension for certain
+code regions, but without change the arch attribute, that means it won't raise
+the minimal execution environment requirement, so user should take care to
+execution the code regions around `.push_arch`/`.pop_arch`, typical use case is
+used with `ifunc`, e.g. libc are built with `rv64gc`, but few functions like
+memcpy provide two version, one built with `rv64gc`, and one built with
+`rv64gcv`, and select by ifunc mechanism at run-time, however we don't want to
+change to minimal execution environment requirement to `rv64gcv`, since
```
jim-wilson, line 197:
```suggestion
change the minimal execution environment requirement to `rv64gcv`, since
```

riscv-asm.md
```diff
@@ -182,6 +182,67 @@ Directive    | Arguments                      | Description
...
+`rv64gcv` version will invoked only if execution environment is supported
```
jim-wilson, line 198:
```suggestion
the `rv64gcv` version will be invoked only if the execution environment supports
```

riscv-asm.md
```diff
@@ -182,6 +182,67 @@ Directive    | Arguments                      | Description
...
+for vector extension, so the minimal execution environment requirement still is
```
jim-wilson, line 199:
```suggestion
the vector extension, so the minimal execution environment requirement still is
```

riscv-asm.md
```diff
@@ -182,6 +182,67 @@ Directive    | Arguments                      | Description
...
+`rv64gc`.
+
+```
+EXTENSION_LIST := EXTENSION_LIST ',' EXTENSION_LIST
```
jim-wilson, line 203:
```suggestion
EXTENSION_LIST := EXTENSION_LIST ',' EXTENSION
```

jim-wilson (approved):
Looks OK with some English improvements.  Only useful if both clang and GNU as support it at roughly the same time, and will require configure checks in GNU tools before using, and whatever the clang equivalent is.

jrtc27:
Why is this a new directive rather than just a `.option`? We already have `.option push`/`.option pop`, all you need is a .option to change the arch string like now we already have rvc/norvc. Perhaps that should just be generalised to (no)rv$arch?

jrtc27:
i.e.:
```
.attribute arch, "rv64imafdc"
memcpy_general:
    add     a5,a1,a2
    beq     a1,a5,.L2
    add     a2,a0,a2
    mv      a5,a0
.L3:
    addi    a1,a1,1
    addi    a5,a5,1
    lbu     a4,-1(a1)
    sb      a4,-1(a5)
    bne     a5,a2,.L3
.L2:
    ret

.option push
.option rvv
memcpy_vec:
    mv a3, a0
.Lloop:
    vsetvli t0, a2, e8, m8, ta, ma
    vle8.v v0, (a1)
    add a1, a1, t0
    sub a2, a2, t0
    vse8.v v0, (a3)
    add a3, a3, t0
    bnez a2, .Lloop
    ret
.option pop    # Restore the enabled ISA extension status to imafdc.
```
This is cleaner, more consistent and avoids adding more than one way to enable/disable RVC.

jrtc27:
> Looks OK with some English improvements. Only useful if both clang and GNU as support it at roughly the same time, and will require configure checks in GNU tools before using, and whatever the clang equivalent is.

The assembler is integrated (well, it doesn't even bother to serialise and re-parse textual assembly) so normally no such concern exists. When there is enough interest in supporting old GNU assemblers with the integrated assembler turned off there's -fbinutils-version that has to be supplied by the user. But configure-time does not make sense for Clang as there's very little distinction between native and cross-compilation beyond the fact it picks a default triple, and every build contains all backends by default, and you're not going to have an assembler for every possible supported target present when you build Clang.

kito-cheng:
> Why is this a new directive rather than just a .option? We already have .option push/.option pop, all you need is a .option to change the arch string like now we already have rvc/norvc. Perhaps that should just be generalised to (no)rv$arch?

Reason why define new directive rather than extend `rvc`/`norvc` scheme.
- Able to included the version info in the option.
- Able to enable more than one extension at a time.
- Implementation is simpler, it won't mixed with existing `.option` implementation.

Here is an another issue there: should we have negative option for `.push_arch`/`.pop_arch`?
e.g. `.push_arch -c`, `.push_arch -v`

jrtc27:
> > Why is this a new directive rather than just a .option? We already have .option push/.option pop, all you need is a .option to change the arch string like now we already have rvc/norvc. Perhaps that should just be generalised to (no)rv$arch?
> 
> Reason why define new directive rather than extend `rvc`/`norvc` scheme.
> 
> * Able to included the version info in the option.

```.option rvv1p0```

> * Able to enable more than one extension at a time.
> * Implementation is simpler, it won't mixed with existing `.option` implementation.
> 
> Here is an another issue there: should we have negative option for `.push_arch`/`.pop_arch`?
> e.g. `.push_arch -c`, `.push_arch -v`

**Comment by jrtc27** [comment]
Other than aesthetics and arguments about simplicity/consistency/cleanliness, there is a very real concern I have about having two different push/pop mechanisms. What does:
```
.attribute arch, rv32i2p0
.push_arch v
.option rvc
.pop_arch
```
mean? Is RVC enabled or disabled at the end of that? Or:
```
.attribute arch, rv32i2p0
.option push
.push_arch v
.option pop
```
Does that have RVV enabled or disabled at the end? If you make the push and pop parts of `.push_arch` and `.pop_arch` be the same as `.option` then they are completely redundant and should not exist beyond the "set the arch string" part (and thus push_arch is a bad name for that part). If you don't make them equivalent and try and maintain two sets of stacks then that becomes a complete tangled nightmare both in the implementation and in the conceptual model. Thus, whatever you name the thing, it needs to ultimately be equivalent to `.option (no)rv$arch`, and so might as well be called that for consistency with (no)rvv rather than some kind of `.modify_arch`.

**Comment by kito-cheng** [comment]
@jrtc27 thanks your example, that's good point, let me write down second version for this proposal.

**Comment by Nelson1225** [comment]
I already have the first version of .push_arch and .pop_arch in GNU binutils, but haven't written the test cases yet.
https://github.com/Nelson1225/riscv-binutils-gdb/commit/8e2483f6a508b908f68b5079fcc65f84ed6381bf

The implementation and assembly syntax of `.option rvc; .option norvc;` and `.push_arch c2p0; .pop_arch` are different, though they all need to be stored in the stack.  I agree with @jrtc27 that we should use the same format and way to do this stuff, so seems that there are some issues as follows,

 1. Extend `.option rvc` that can have versions.  for example,
`.attribute arch, rv32i2p0`
 2. ...

**Comment by Nelson1225** [comment]
I already have the first version of .push_arch and .pop_arch in GNU binutils, but haven't written the test cases yet.
Nelson1225/riscv-binutils-gdb@8e2483f

The implementation and assembly syntax of .option rvc; .option norvc; and .push_arch c2p0; .pop_arch are different, though they all need to be stored in the stack. I agree with @jrtc27 that we should use the same format and way to do this stuff, so seems that there are some issues as follows,

1. Extend .option rvc that can have versions. for example,
```
.attribute arch, rv32i2p0
.option push  /* backup rv32i2p0.  */
.option rvc2p0  /* added c2p0, rv32i2p0_c2p0.  */
.option rvc3p0  /* updated c3p0, rv32i2p0_c3p0.  */
.option norvc  /* removed c, rv32i2p0.  */
.option pop  /* restored rv32i2p0.  */

```
So the format `.option rv<extension><version>` is good, but `.option norv<extension><version>` is hard to use.

2. Not only the standard extensions can be pushed and popped, so according to the above formats.
```
.option rv<extension><version> and .option norv<extension>

.attribute arch, rv32i2p0_m2p0
.option push  /* backup rv32i2p0.  */
.option rvc2p0  /* added c2p0, rv32i2p0_c2p0.  */
.option rvzfh0p1  /* added zfh0p1, rv32i2p0_c2p0_zfh0p1.  */
.option rvxvendor1p0  /* added xvendor1p0, rv32i2p0_c2p0_zfh0p1_xvendor1p0.  */
.option pop  /* restored rv32i2p0.  */
```
Although `rv` prefix looks redundant, but we need to consider the compatibility of rvc and norvc.

3. Should we still support the following format?
```
.option rv<extension1><versionN>, rv<extension2><versionM>, ...
.option norv<extension1>, norv<extension2>, ...
```

**Comment by kito-cheng** [comment]
Changes:
- Extend `.option` rather than `.push_arch`/`.pop_arch`
- Thanks @Nelson1225 implement PoC for this proposal!
   -  https://sourceware.org/pipermail/binutils/2021-July/117531.html

**Comment by jrtc27** [comment]
Why is this still introducing a new way of doing things, `.option arch <foo>`, rather than reusing and extending the existing syntax? This _still_ results in two different ways to enable/disable RVC. We should not be adding new syntax unless _absolutely necessary_.

File: riscv-asm.md [hunk 5]
```diff
@@ -183,6 +183,116 @@ Directive    | Arguments                      | Description
...
+without changing the arch attribute, that means it won't raise the minimal
+execution environment requirement, so the user should take care to protect
+execution of the code regions around `.option push`/`.option arch`/`.option pop`.
+A typical use case is is with `ifunc`, e.g. the libc is built with `rv64gc`,
+but a few functions like memcpy provide two versions, one built with `rv64gc`
+and one built with `rv64gcv`, and then select between them by ifunc mechanism
+at run-time.  However we don't want to change the minimal execution environment
+requirement to `rv64gcv`, since the `rv64gcv` version will be invoked only if
+the execution environment supports the vector extension, so the minimal
+execution environment requirement still is `rv64gc`.
```
palmer-dabbelt, line 206: This should be commentary.

palmer-dabbelt:
IMO this is the right way to go: having `.option arch +c` mean the same thing as `.option rvc` is fine, it's just a more general way of doing things.  It's way cleaner than trying to have a million ad-hoc '.option noZba` for every flavor of extension we want to flip on/off.  I'd go a bit farther and suggest a handful of additions:

* Add an `=` op (or just let it be blank).  This would let users write something like `.option arch rv64gc` when they want to control exactly the arch they're based on.
* Drop the extension list, and just make users insert multiple `.option arch` lines.  This prevents us from getting screwed when our separator (IIUC a `,`?) becomes a valid character in extensions.
* Add this as command-line syntax.  Something like `-march=rv64g -march+c`.  We can deal with that later, though, as I'm sure it'll require a lot of sorting out.
* Add this as a function attribute.  So something like `int func() __attribute__(("arch", "-c"))` or whatever the syntax usually looks like over there.  We can also argue about this one later.

jrtc27:
> IMO this is the right way to go: having `.option arch +c` mean the same thing as `.option rvc` is fine, it's just a more general way of doing things. It's way cleaner than trying to have a million ad-hoc '.option noZba` for every flavor of extension we want to flip on/off. I'd go a bit farther and suggest a handful of additions:

Just as you wouldn't implement this by having a million ad-hoc -Zba options for .option arch, you wouldn't implement my proposal that way either, you'd parse (no)rv and then parse an extension name.

> * Add an `=` op (or just let it be blank).  This would let users write something like `.option arch rv64gc` when they want to control exactly the arch they're based on.
> * Drop the extension list, and just make users insert multiple `.option arch` lines.  This prevents us from getting screwed when our separator (IIUC a `,`?) becomes a valid character in extensions.
> * Add this as command-line syntax.  Something like `-march=rv64g -march+c`.  We can deal with that later, though, as I'm sure it'll require a lot of sorting out.
> * Add this as a function attribute.  So something like `int func() __attribute__(("arch", "-c"))` or whatever the syntax usually looks like over there.  We can also argue about this one later.

That's `__attribute__((target(...)))` as an architecture-independent attribute (with architecture-dependent arguments), and `__attribute__((target("-c")))` etc already work today for LLVM (in as much as it can without assembly syntax to toggle things on and off; for C in particular it won't work well, or at least not if you round-trip through assembly, might work fine if you don't, but for something like M or F it'll just (not) emit the instructions).

kito-cheng:
>> IMO this is the right way to go: having .option arch +c mean the same thing as .option rvc is fine, it's just a more general way of doing things. It's way cleaner than trying to have a million ad-hoc '.option noZba` for every flavor of extension we want to flip on/off. I'd go a bit farther and suggest a handful of additions:
>
> Just as you wouldn't implement this by having a million ad-hoc -Zba options for .option arch, you wouldn't implement my proposal that way either, you'd parse (no)rv and then parse an extension name.

To me, both are extending stuffs, and now we have chance to implement one scheme more clear / better, so I would prefer the new one. (although that's subjective I know.)

kito-cheng:
> Add an = op (or just let it be blank). This would let users write something like .option arch rv64gc when they want to control exactly the arch they're based on.

Sound good idea.

> Drop the extension list, and just make users insert multiple .option arch lines. This prevents us from getting screwed when our separator (IIUC a ,?) becomes a valid character in extensions.

Hmmm, make sense to me... 

> Add this as command-line syntax. Something like -march=rv64g -march+c. We can deal with that later, though, as I'm sure it'll require a lot of sorting out.

Yeah, I guess we need that to deal with the RISC-V `profile`...although I prefer use -march as the only option for control the ISA.

> Add this as a function attribute. So something like int func() __attribute__(("arch", "-c")) or whatever the syntax usually looks like over there. We can also argue about this one later.

Like @jrtc27 said, LLVM having some level of support for that, write down that into to https://github.com/riscv/riscv-c-api-doc is my next step, but for GNU toolchain site, we need this to implement that.

palmer-dabbelt:
> > IMO this is the right way to go: having `.option arch +c` mean the same thing as `.option rvc` is fine, it's just a more general way of doing things. It's way cleaner than trying to have a million ad-hoc '.option noZba` for every flavor of extension we want to flip on/off. I'd go a bit farther and suggest a handful of additions:
> 
> Just as you wouldn't implement this by having a million ad-hoc -Zba options for .option arch, you wouldn't implement my proposal that way either, you'd parse (no)rv and then parse an extension name.

IMO that's pretty ad-hoc: by removing the explicit name-spacing provided by Kito's proposal we end up with these options all at the top level, which makes them harder to describe.  That said, I don't really care that much about syntax.

> > * Add an `=` op (or just let it be blank).  This would let users write something like `.option arch rv64gc` when they want to control exactly the arch they're based on.
> > * Drop the extension list, and just make users insert multiple `.option arch` lines.  This prevents us from getting screwed when our separator (IIUC a `,`?) becomes a valid character in extensions.
> > * Add this as command-line syntax.  Something like `-march=rv64g -march+c`.  We can deal with that later, though, as I'm sure it'll require a lot of sorting out.
> > * Add this as a function attribute.  So something like `int func() __attribute__(("arch", "-c"))` or whatever the syntax usually looks like over there.  We can also argue about this one later.
> 
> That's `__attribute__((target(...)))` as an architecture-independent attribute (with architecture-dependent arguments), and `__attribute__((target("-c")))` etc already work today for LLVM (in as much as it can without assembly syntax to toggle things on and off; for C in particular it won't work well, or at least not if you round-trip through assembly, might work fine if you don't, but for something like M or F it'll just (not) emit the instructions).

Sounds like that should be in some spec somewhere, as I don't remember having seen it before.  I also very much question the value of providing options that generate broken code.

jrtc27:
> > > IMO this is the right way to go: having `.option arch +c` mean the same thing as `.option rvc` is fine, it's just a more general way of doing things. It's way cleaner than trying to have a million ad-hoc '.option noZba` for every flavor of extension we want to flip on/off. I'd go a bit farther and suggest a handful of additions:
> > 
> > 
> > Just as you wouldn't implement this by having a million ad-hoc -Zba options for .option arch, you wouldn't implement my proposal that way either, you'd parse (no)rv and then parse an extension name.
> 
> IMO that's pretty ad-hoc: by removing the explicit name-spacing provided by Kito's proposal we end up with these options all at the top level, which makes them harder to describe. That said, I don't really care that much about syntax.
> 
> > > * Add an `=` op (or just let it be blank).  This would let users write something like `.option arch rv64gc` when they want to control exactly the arch they're based on.
> > > * Drop the extension list, and just make users insert multiple `.option arch` lines.  This prevents us from getting screwed when our separator (IIUC a `,`?) becomes a valid character in extensions.
> > > * Add this as command-line syntax.  Something like `-march=rv64g -march+c`.  We can deal with that later, though, as I'm sure it'll require a lot of sorting out.
> > > * Add this as a function attribute.  So something like `int func() __attribute__(("arch", "-c"))` or whatever the syntax usually looks like over there.  We can also argue about this one later.
> > 
> > 
> > That's `__attribute__((target(...)))` as an architecture-independent attribute (with architecture-dependent arguments), and `__attribute__((target("-c")))` etc already work today for LLVM (in as much as it can without assembly syntax to toggle things on and off; for C in particular it won't work well, or at least not if you round-trip through assembly, might work fine if you don't, but for something like M or F it'll just (not) emit the instructions).
> 
> Sounds like that should be in some spec somewhere, as I don't remember having seen it before. I also very much question the value of providing options that generate broken code.

https://clang.llvm.org/docs/AttributeReference.html#target
https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html (scroll way down)

It works just fine for everything but C. For C it'll still mess with the subtarget internally (which can affect codegen for relative costs of instructions as it looks at whether they're compressible) but we fail to omit the necessary `.option (no)rvc` to stop post-isel compression, and it seems that does affect using the integrated assembler too (not sure how though, bit strange). There's nothing wrong with the option, just a bug in its implementation for one specific case (that's a rather weird one).

jrtc27:
> > > IMO this is the right way to go: having `.option arch +c` mean the same thing as `.option rvc` is fine, it's just a more general way of doing things. It's way cleaner than trying to have a million ad-hoc '.option noZba` for every flavor of extension we want to flip on/off. I'd go a bit farther and suggest a handful of additions:
> > 
> > 
> > Just as you wouldn't implement this by having a million ad-hoc -Zba options for .option arch, you wouldn't implement my proposal that way either, you'd parse (no)rv and then parse an extension name.
> 
> IMO that's pretty ad-hoc: by removing the explicit name-spacing provided by Kito's proposal we end up with these options all at the top level, which makes them harder to describe. That said, I don't really care that much about syntax.

I agree that, if I could go back in time, I would stop the (no)rvc syntax from being adopted by GNU as and pick something more like this proposal. However, my personal opinion is that there should be a very good reason for introducing new syntax for something that can be done by generalising existing syntax, but `.option arch` is certainly better than `.push_arch`. I think the way forward is to bring it up with both GNU and LLVM developers to see if there's a consensus for which of the two current proposals they feel is best.

jrtc27:
NB: I would personally put a comma after `.option arch`, as is done for `.attribute arch`.

kito-cheng:
Changes:
- Add `,` after arch.
- Only allow enable/disable one arch one time.
- Add `=` operator to set arch to specific configuration.

asb:
I agree that `.option arch ` is just a more general way to specify things, and I prefer keeping things namespaced there.

We discussed somewhat on the LLVM sync-up call and there were diverging opinions on whether it was desirable to later deprecate `.option [no]rvc` (either just documenting it's no longer the preferred way, or even moving towards emitting warnings). But I don't think that's a discussion that needs to be had here and now.

kito-cheng:
Changes:
- Fix wrong syntax in `rvc/norvc` section

riscv-asm.md
luismarques: If this is changed to commentary, it's important that the part about "won't raise the minimal execution environment requirement" be left as normative. We don't want to set `EF_RISCV_RVC` in the ELF flags when using `.option arch, +c`, even though binutils currently does that for `.option rvc`. See https://reviews.llvm.org/D122490#3419774 for a recent discussion about that.

kito-cheng:
Changes:
- Move `ifunc` example to note.
- Mention `EF_RISCV_RVC` won't updated by `.option arch`
- Claim we'll deprecate `rvc`/`norvc` in future.
- Add more note for `rvc`/`norvc`  option.

luxufan:
Hi, @kito-cheng . I have a question that since D extension will imply F extension, should '.option arch, -d' also disable F extension at the same time?

jrtc27:
Why wouldn't that just leave you with F? The only weird thing is that `.option arch, =rv32i, +d, -d` will give you rv32if, but I think that's the least bad semantics and hard to oddities like that.

kito-cheng:
> I have a question that since D extension will imply F extension, should '.option arch, -d' also disable F extension at the same time?

No, since that means we also need to disable zicsr due to F imply zicsr :(
But thanks for point out that I think we should add NOTE to mention that.

> The only weird thing is that .option arch, =rv32i, +d, -d will give you rv32if, but I think that's the least bad semantics and hard to oddities like that.

@jrtc27 I guess we should add a NOTE to recommend user use `push`/`pop` rather than `-`, consider =rv32i, +v, -v, will result `rv32ifd_zve32x_zve32f_zve64x_zve64f_zve64d_zvl32b_zvl64b_zvl128b`.

kito-cheng:
Changes:
- Rebase
- Add NOTE to address @luxufan's and @jrtc27's comment

riscv-asm.md
```diff
@@ -184,6 +184,150 @@ Directive    | Arguments                      | Description
...
+Enable/disable the C-extension for the following code region, this option is equivalent
+to `.option arch, +c`/`.option arch, -c`, but widely support by older version of
+toolchain.
+
+NOTE: `.option rvc` might set the `EF_RISCV_RVC` in the ELF flags on specific
+toolchains, that might cause linker convert instruction into compressed
+instruction on unexpected region.
+
+NOTE: The difference between `.option rvc`/`.option norvc` and
+`.option arch +c`/`.option arch -c` won't set `EF_RISCV_RVC` in the ELF flags.
```
luismarques, line 203: NOTE: There is a difference between `.option rvc`/`.option norvc` and
`.option arch +c`/`.option arch -c`. The latter won't set `EF_RISCV_RVC` in the ELF flags.

riscv-asm.md
```diff
@@ -184,6 +184,150 @@ Directive    | Arguments                      | Description
...
+
+#### `arch`
+
+Enable and/or disable specific ISA extensions for following code regions, but
```
luismarques, line 207: "for following" -> "for the following"

riscv-asm.md
```diff
@@ -184,6 +184,150 @@ Directive    | Arguments                      | Description
...
+
+Set the code model to PIC (position independent code) or non-PIC. This will
+affect the expansion of the `la` pseudoinstruction, refer to
+[listing of standard RISC-V pseudoinstructions](#pseudoinstructions).
+
+#### `relax`/`norelax`
+
+Enable/disable linker relaxation for the following code region.
+
+NOTE: Code region follows by `.option relax` will emit
```
luismarques, line 322: "follows by" -> "followed by"

riscv-asm.md
```diff
@@ -184,6 +184,150 @@ Directive    | Arguments                      | Description
...
+`R_RISCV_RELAX`/`R_RISCV_ALIGN` even linker unsupport relaxation, suggested
```
luismarques, line 323: "even linker unsupport relaxation" -> "even if the linker does not support relaxation"?

kito-cheng:
Changes:
- Rebase
- Apply @luismarques's suggestion

kito-cheng:
@luismarques thanks!

kito-cheng:
@luismarques @asb do you mind take a review for this? this has been verified on binutils for a while and here also a LLVM implementation already under review: https://reviews.llvm.org/D123515

riscv-asm.md
```diff
@@ -188,7 +188,128 @@ Directive    | Arguments                      | Description
...
+```
+
+.option arch, <EXTENSION>
+
+OP             := '+'
+                | '-'
+
+EXTENSION      := OP <extension-name> VERSION
+                | OP <extension-name>
+                | '=' <full-arch-string>
```
jrtc27, line 225: binutils dropped the leading =

riscv-asm.md
```diff
@@ -188,7 +188,128 @@ Directive    | Arguments                      | Description
...
+Enable and/or disable specific ISA extensions for the following code regions, but
+without changing the arch attribute and `EF_RISCV_RVC` in the ELF flags, that
+means it won't raise the minimal execution environment requirement, so the user
+should take care to protect execution of the code regions around
+`.option push`/`.option arch`/`.option pop`.
+
+Syntax for `.option arch`:
+```
+
+.option arch, <EXTENSION>
```
jrtc27, line 218: binutils has:
```
.option arch, <EXTENSION-LIST>

EXTENSION-LIST := <EXTENSION>
                | <EXTENSION>, <EXTENSION-LIST>
```

riscv-asm.md
```diff
@@ -188,7 +188,128 @@ Directive    | Arguments                      | Description
...
+`rv32i` + `.option arch, +v` will also enable `f`, `d`, `zve32x`, `zve32f`,
+`zve64x`, `zve64f`, `zve64d`, `zvl32b`, `zvl64b` and `zvl128b` extensions.
+
+NOTE: We recommended `.option arch, +` and `.option arch, -` are used with
+`.option push`/`.option pop` instead of a `.option arch, +` / `.option arch, -`
+pair, because `.option arch, +` will enable all required extensions, but
+`.option arch, -` only disable specific extension, so the result might be
+unexpected, for example: `rv32i` + `.option arch, +v` + `.option arch, -v`
+will result `rv32ifd_zve32x_zve32f_zve64x_zve64f_zve64d_zvl32b_zvl64b_zvl128b`
+not `rv32i`.
```
jrtc27, line 312: More clarity is needed for:
```
.option arch, rv64ifd
.option arch, -f
```
Currently binutils will treat that that as a no-op, since it will turn it into rv64id and then immediately re-compute the implications. I view this as surprising and not that helpful; ideally it would result in rv64i instead. Otherwise you need to know the full list of possible extensions and list all the ones _that the current toolchain supports_, which will change over time as new extensions get added (like how parts of V and Zfh were added after F).

riscv-asm.md
```diff
@@ -188,7 +188,128 @@ Directive    | Arguments                      | Description
...
+older toolchain versions.
+
+Alternative style:
+NOTE: .option rvc might set the ELF flag EF_RISCV_RVC in some toolchains. That
+might cause the linker to compress instructions in code regions where that was
+not intended.
+
+NOTE: There is a difference between `.option rvc`/`.option norvc` and
+`.option arch +c`/`.option arch -c`. The latter won't set EF_RISCV_RVC in the
+ELF flags.
```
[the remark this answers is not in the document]
kito-cheng: `EF_RISCV_RVC` should be deprecated IMO, linker currently rely on that to decide linker relaxation can use C instruction or not, but that apparently not a completeness solution.

Give an example to show what's wrong:
```
.attribute arch, "rv64g" # NO RVC!

...
# Only allow rvc in following region
.option rvc
...
.option norvc
...
```

However option rvc will set `EF_RISCV_RVC` will cause whole program might relaxed into C instruction even not in the rvc/norvc region.

riscv-asm.md
kito-cheng: Add example for this case.

riscv-asm.md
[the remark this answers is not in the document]
Nelson1225: Agreed.  FYI, there are two patches might help to clarify the idea,

* RISC-V: File-level architecture shouldn't be affected by section-level ones
https://sourceware.org/pipermail/binutils/2022-November/124178.html

* RISC-V: Refer mapping symbol to R_RISCV_RELAX for rvc relaxations
https://sourceware.org/pipermail/binutils/2022-September/123191.html

kito-cheng:
We've binutils support on upstream and a patch on LLVM (https://reviews.llvm.org/D123515), so I think it's time to merge this, before merge this, I would like wait one more week to make sure no objection or further comment from LLVM community @asb @luismarques @jrtc27 @MaskRay.

cmuellner (approved):

riscv-asm.md
```diff
@@ -214,7 +214,150 @@ ATTRIBUTE_VALUE := <string>
...
+.option push
+.option +c   # Alternative of .option rvc
+.option pop
+
+.option push
+.option -c   # Alternative of .option norvc
+.option pop
+```
+
+NOTE: .option rvc might set the ELF flag EF_RISCV_RVC in some toolchains. That
```
MaskRay, line 236: backticks around `.option rvc` and `EF_RISCV_RVC`.

riscv-asm.md
```diff
@@ -214,7 +214,150 @@ ATTRIBUTE_VALUE := <string>
...
+.L3:
+    addi    a1,a1,1
+    addi    a5,a5,1
+    lbu     a4,-1(a1)
+    sb      a4,-1(a5)
+    bne     a5,a2,.L3
+.L2:
+    ret
+
+.option push     # Push current option to stack.
```
MaskRay, line 300: Push current options to the stack.

riscv-asm.md
```diff
@@ -214,7 +214,150 @@ ATTRIBUTE_VALUE := <string>
...
+    addi    a1,a1,1
+    addi    a5,a5,1
+    lbu     a4,-1(a1)
+    sb      a4,-1(a5)
+    bne     a5,a2,.L3
+.L2:
+    ret
+.option pop   # Pop current option from stack, restore the enabled ISA extension status to imafdc.
+
+.option push           # Push current option to stack.
```
MaskRay, line 332: misaligned `#`

riscv-asm.md
```diff
@@ -214,7 +214,150 @@ ATTRIBUTE_VALUE := <string>
...
+VERSION                := [0-9]+ 'p' [0-9]+
+                        | [1-9][0-9]*
+
+EXTENSION-NAME         := Naming rule is defined in RISC-V ISA manual
+```
+
+- Extension version can be omitted, the assembler will use the built-in default
+  version for that extension.
+- `OP` can be enable (`+`) or disable (`-`).
+- Input format of `<full-arch-string>` is the same as `-march` option.
```
MaskRay, line 279: The format of

I think `input` can be omitted.

riscv-asm.md
```diff
@@ -214,7 +214,150 @@ ATTRIBUTE_VALUE := <string>
...
+
+Example:
+
+```assembly
+.attribute arch, "rv64imafdc"
+# You can only use instruction from i, m, a, f, d and c extensions.
```
MaskRay, line 285: instructions from the

MaskRay (approved):

riscv-asm.md
```diff
@@ -214,7 +214,150 @@ ATTRIBUTE_VALUE := <string>
...
+.option -c   # Alternative of .option norvc
+.option pop
+```
+
+NOTE: .option rvc might set the ELF flag EF_RISCV_RVC in some toolchains. That
+might cause the linker to compress instructions in code regions where that was
+not intended.
+
+NOTE: There is a difference between `.option rvc`/`.option norvc` and
+`.option arch +c`/`.option arch -c`. The latter won't set EF_RISCV_RVC in the
```
MaskRay, line 241: `.option arch, +c`

Better not to allow `.option .arch +c` (space-separated).

riscv-asm.md
```diff
@@ -214,7 +214,150 @@ ATTRIBUTE_VALUE := <string>
...
+
+Enable/disable the C-extension for the following code region. This option is
+equivalent to `.option arch, +c`/`.option arch, -c`, but widely support by
+older toolchain versions.
+
+Alternative style:
+
+```
+.option push
+.option +c   # Alternative of .option rvc
```
MaskRay, line 228: `.option arch, +c`

I think you do not intend to allow `.option +c`, right?

riscv-asm.md
```diff
@@ -214,7 +214,150 @@ ATTRIBUTE_VALUE := <string>
...
+
+FULLARCHSTR            := <full-arch-string>
+
+EXTENSION              := <OP> <EXTENSION-NAME> [<VERSION>]
+
+OP                     := '+'
+                        | '-'
+
+VERSION                := [0-9]+ 'p' [0-9]+
+                        | [1-9][0-9]*
```
luismarques, line 271: I'm not sure this is correct. Can you please walk me through it?

riscv-asm.md
```diff
@@ -214,7 +214,150 @@ ATTRIBUTE_VALUE := <string>
...
+
+- Extension version can be omitted, the assembler will use the built-in default
+  version for that extension.
+- `OP` can be enable (`+`) or disable (`-`).
+- Input format of `<full-arch-string>` is the same as `-march` option.
+
+Example:
+
+```assembly
+.attribute arch, "rv64imafdc"
```
luismarques, line 284: Update to match binutils, remove quotes.

riscv-asm.md
```diff
@@ -214,7 +214,150 @@ ATTRIBUTE_VALUE := <string>
...
+mechanism at run-time.  However, we don't want to change the minimal execution
+environment requirement to `rv64gcv`, since the `rv64gcv` version will be
+invoked only if the execution environment supports the vector extension, so
+the minimal execution environment requirement still is `rv64gc`.
+
+NOTE: `.option arch, +` will also enable all required extensions, for example,
+`rv32i` + `.option arch, +v` will also enable `f`, `d`, `zve32x`, `zve32f`,
+`zve64x`, `zve64f`, `zve64d`, `zvl32b`, `zvl64b` and `zvl128b` extensions.
+
+NOTE: We recommended `.option arch, +` and `.option arch, -` are used with
```
luismarques, line 351:
```suggestion
NOTE: We recommend `.option arch, +` and `.option arch, -` are used with
```

riscv-asm.md
```diff
@@ -214,7 +214,150 @@ ATTRIBUTE_VALUE := <string>
...
+`rv64gc`, but a few functions like memcpy provide two versions, one built with
+`rv64gc` and one built with `rv64gcv`, and then select between them by ifunc
+mechanism at run-time.  However, we don't want to change the minimal execution
+environment requirement to `rv64gcv`, since the `rv64gcv` version will be
+invoked only if the execution environment supports the vector extension, so
+the minimal execution environment requirement still is `rv64gc`.
+
+NOTE: `.option arch, +` will also enable all required extensions, for example,
+`rv32i` + `.option arch, +v` will also enable `f`, `d`, `zve32x`, `zve32f`,
+`zve64x`, `zve64f`, `zve64d`, `zvl32b`, `zvl64b` and `zvl128b` extensions.
```
luismarques, line 349: It would be good to explain somewhere that e.g. `+c, -c` and `+v, -v` do the expected things. I guess the expected thing is that the named extension is disabled but its dependencies are enabled.

luismarques (changes requested):

riscv-asm.md
luismarques: (alternatively it could be an error)

riscv-asm.md
kito-cheng: I guess the current rule might be confused by `[`, `]`, so I try to rewrite the rule to allow VERSION as empty string, also expand first rule into two for forbid `0p0`.

```
EXTENSION              := <OP> <EXTENSION-NAME> [<VERSION>]

VERSION                := [0-9]+ 'p' [0-9]+
                        | [1-9][0-9]*
```

to

```
EXTENSION              := <OP> <EXTENSION-NAME> <VERSION>

VERSION                := [1-9][0-9]* 'p' [0-9]+
                        | [0-9]+ 'p' [1-9][0-9]*
                        | [1-9][0-9]*
                        |
```

riscv-asm.md
kito-cheng: Fixed

riscv-asm.md
kito-cheng: Updated, that's typo, I am not intend to support space-separated form

riscv-asm.md
kito-cheng: @luismarques let me know it's still unclear to you or any case you think it's still ambiguous

riscv-asm.md
```diff
@@ -359,6 +359,10 @@ Another example is `.option arch, rv64ifd` + `.option arch, -f` will still got
 same result `rv64ifd`, because `f` will add back during compute implication
 extensions for `rv64id`.
 
+NOTE: `.option arch, +<ext>, -<ext>` can be used to enable dependency
+extensions, e.g. `rv32i` + `.option arch, +v, -v` will result
+`rv32ifd_zve32x_zve32f_zve64x_zve64f_zve64d_zvl32b_zvl64b_zvl128b`.
```
luismarques, line 364:
```suggestion
NOTE: `.option arch, +<ext>, -<ext>` is accepted and will result in enabling the
extensions that depend on `ext`, e.g. `rv32i` + `.option arch, +v, -v` will result
`rv32ifd_zve32x_zve32f_zve64x_zve64f_zve64d_zvl32b_zvl64b_zvl128b`.
```

luismarques (approved):
LGTM. Thanks for all this effort Kito!

riscv-asm.md
luismarques: LGTM now.

PR #68, opened 2021-07-27
Add more description for .option directive

kito-cheng:


riscv-asm.md
```diff
@@ -183,6 +183,25 @@ Directive    | Arguments                      | Description
...
+### <a name=.option></a> `.option`
+
+#### `rvc`/`norvc`
+
+Enable/disable C-extension for following code region.
+
+#### `pic`/`nopic`
+
+Set the code mode is PIC (position independent code) or non-PIC mode, this will
+affect how `la` expand.
```
jim-wilson, line 195:
```suggestion
Set the code mode to PIC (position independent code) or non-PIC mode, this will
affect how `la` expands.
```

jim-wilson (approved):
Some linkers don't support relaxation, but that probably doesn't need to be explained here.  if it does, then maybe something like "if the linker supports it" added to the relax explanation would be good enough.

kito-cheng:
Changes:
- Apply Jim's suggestions

riscv-asm.md
```diff
@@ -164,7 +164,7 @@ Directive    | Arguments                      | Description
 .macro       | name arg1 [, argn]             | begin macro definition \argname to substitute
 .endm        |                                | end macro definition
 .type        | symbol, @function              | accepted for source compatibility
-.option      | {rvc,norvc,pic,nopic,push,pop} | RISC-V options
+.option      | {rvc,norvc,pic,nopic,relax,norelax,push,pop} | RISC-V options, more detailed description see [.optino](#.option).
```
jrtc27, line 167:
```suggestion
.option      | {rvc,norvc,pic,nopic,relax,norelax,push,pop} | RISC-V options, more detailed description in [.option](#.option).
```

riscv-asm.md
```diff
@@ -183,6 +183,26 @@ Directive    | Arguments                      | Description
...
 
+### <a name=.option></a> `.option`
+
+#### `rvc`/`norvc`
+
+Enable/disable C-extension for following code region.
+
+#### `pic`/`nopic`
+
+Set the code mode to PIC (position independent code) or non-PIC mode, this will
```
jrtc27, line 194:
```suggestion
Set the code model to PIC (position independent code) or non-PIC, this will
```

riscv-asm.md
```diff
@@ -183,6 +183,26 @@ Directive    | Arguments                      | Description
 .balign      | b,[pad_val=0]                  | byte align
 .zero        | integer                        | zero bytes
 
+### <a name=.option></a> `.option`
+
+#### `rvc`/`norvc`
+
+Enable/disable C-extension for following code region.
```
jrtc27, line 190:
```suggestion
Enable/disable C-extension for the following code region.
```

riscv-asm.md
```diff
@@ -183,6 +183,26 @@ Directive    | Arguments                      | Description
...
+
+#### `pic`/`nopic`
+
+Set the code mode to PIC (position independent code) or non-PIC mode, this will
+affect how `la` expands.
+
+#### `relax`/`norelax`
+
+Enable/disable linker relaxation for following code region if the linker
```
jrtc27, line 199:
```suggestion
Enable/disable linker relaxation for the following code region if the linker
```

kito-cheng:
@jrtc27 does this change look good to you after apply your revision?

kito-cheng:
@luismarques @asb @jrtc27 did you mind did a final round review?

riscv-asm.md
```diff
@@ -183,6 +183,26 @@ Directive    | Arguments                      | Description
...
+
+#### `pic`/`nopic`
+
+Set the code model to PIC (position independent code) or non-PIC, this will
+affect how `la` expands.
+
+#### `relax`/`norelax`
+
+Enable/disable linker relaxation for the following code region if the linker
+supports it.
```
luismarques, line 200: "if the linker supports it" -> I think this enables it unconditionally, no? At least Clang will just enable it, and if the linker doesn't support it (LLD...) then it craps out.

I don't know if it's worth mentioning that this also includes emission of`R_RISCV_ALIGN` relocations?

riscv-asm.md
```diff
@@ -164,7 +164,7 @@ Directive    | Arguments                      | Description
 .macro       | name arg1 [, argn]             | begin macro definition \argname to substitute
 .endm        |                                | end macro definition
 .type        | symbol, @function              | accepted for source compatibility
-.option      | {rvc,norvc,pic,nopic,push,pop} | RISC-V options
+.option      | {rvc,norvc,pic,nopic,relax,norelax,push,pop} | RISC-V options, more detailed description in [.option](#.option).
```
luismarques, line 167: Reformat table, to realign the `|`s?

Nit: "RISC-V options. Refer to [.option](#.option) for a more detailed description."

riscv-asm.md
```diff
@@ -183,6 +183,26 @@ Directive    | Arguments                      | Description
 .balign      | b,[pad_val=0]                  | byte align
 .zero        | integer                        | zero bytes
 
+### <a name=.option></a> `.option`
+
+#### `rvc`/`norvc`
+
+Enable/disable C-extension for the following code region.
```
luismarques, line 190: "C-extension" -> "The C extension"

I don't know if the word "region" in "the following code region" kind of implies that this has limited scope, like an `#if ... #endif` region. I'll let the native speakers nitpick that :)

riscv-asm.md
```diff
@@ -183,6 +183,26 @@ Directive    | Arguments                      | Description
...
+
+#### `pic`/`nopic`
+
+Set the code model to PIC (position independent code) or non-PIC, this will
+affect how `la` expands.
```
luismarques, line 195: Nit: "Set the code model to PIC (position independent code) or non-PIC. This will affect the expansion of the `la` pseudoinstruction."
Is there some place we can link to that details how `la` is expanded with the various models?

luismarques:
> Some linkers don't support relaxation, but that probably doesn't need to be explained here. if it does, then maybe something like "if the linker supports it" added to the relax explanation would be good enough.

I hadn't noticed your earlier comment, Jim. I think that phrasing ends up being misleading. When using this option the compiler will emit relaxation relocations without checking for linker support, and then the linker will complain if it doesn't actually support such relaxations. Maybe just state that the linker needs to support relaxations?

riscv-asm.md
kito-cheng: @aswaterman did you having some suggestion on the wording about `region`?

riscv-asm.md
kito-cheng: I would prefer don't realign this time since we might need to realign for `.option arch,` soon, my thought is we can did that when converting the format to AsciiDoc.

kito-cheng:
Changes:
- Rebase
- Address @luismarques's comment
- Add NOTE to `.option relax`/`.option norelax`

kito-cheng:
Changes:
- Add one more NOTE to recommend user should use `.option push` + `.option norelax` + `.option pop`  if they want to disable linker relaxation of specific code region

riscv-asm.md
kito-cheng: Add reference to pseudoinstructions

asb:
> @luismarques @asb @jrtc27 did you mind did a final round review?

No additional comments from me.

luismarques:
I don't think I have permission in this repo/org to approve anything, but the technical issues I pointed out seem to have been addressed, so it LGTM.

kito-cheng:
@luismarques @asb thanks you guys, I gonna merge this :)

PR #69, opened 2021-07-30
Document lla, fix table documentation for la and add lga pseudo instruction

jrtc27:


riscv-asm.md
```diff
@@ -540,7 +568,10 @@ fail_msg:
 
 Pseudoinstruction            | Base Instruction(s)                                           | Meaning   | Comment
 :----------------------------|:--------------------------------------------------------------|:----------|:--------|
-la rd, symbol                | auipc rd, symbol[31:12]; addi rd, rd, symbol[11:0]            | Load address
+la rd, symbol                | auipc rd, symbol[31:12]; addi rd, rd, symbol[11:0]            | Load address | Without -fPIC
+la rd, symbol                | auipc rd, symbol@GOT[31:12]; l{w\|d} rd, symbol@GOT[11:0]\(rd\) | Load address | With -fPIC
```
jrtc27, line 572: Wasn't sure what syntax to use here; we're already sloppy about the PC-relative nature of symbol above, and not using proper assembly syntax, so I figured something like this was most in keeping for giving the high-level idea without the specifics that would make the table even more unwieldy.

riscv-asm.md
```diff
@@ -540,7 +568,10 @@ fail_msg:
...
+lla rd, symbol               | auipc rd, symbol[31:12]; addi rd, rd, symbol[11:0]            | Load local address
+lga rd, symbol               | auipc rd, symbol@GOT[31:12]; l{w\|d} rd, symbol@GOT[11:0]\(rd\) | Load global address
```
MaskRay, line 574: The syntax is weird.
In assembly, it's `auipd rd, %got_pcrel_hi(symbol)`

MaskRay (commented):
I agree that a pseudo instruction will be useful.

riscv-asm.md
jrtc27: Yeah, see https://github.com/riscv/riscv-asm-manual/pull/69#discussion_r680192496 for why I went with this

MaskRay:
LGTM

jim-wilson:
One thing I noticed while looking at this is that GNU as handles la and lla exactly the same as li when the symbol is actually a constant.  That can happen if the symbol is set or equated to a constant.  This isn't mentioned in the manual, but I think this is a minor point.

I wrote and submitted a GNU as patch to implement lga.
https://sourceware.org/pipermail/binutils/2021-August/117637.html
to be installed if/when this gets accepted.

jim-wilson (approved):

jrtc27:
That's probably fine. LLVM will reject such cases, only allowing you to `li` something you `.set`; I don't know if that's a good thing or not, but it's _a_ sensible implementation, if not necessarily the most helpful.

riscv-asm.md
ebahapo: Methinks that the canonical pseudo functions should be used instead in order to avoid confusion to users.

PR #71, opened 2021-09-16
Document addi immediate

nick-knight:
The RISC-V assembly language allows negative `addi` immediates to be expressed in more than one way, which may be surprising to newcomers. This PR is intended to resolve riscv-non-isa/riscv-asm-manual#70.

### DIFF
--- a/riscv-asm.md
+++ b/riscv-asm.md
@@ -312,11 +312,13 @@ Load Upper Immediate's Immediate
 The immediate argument to `lui` is an integer in the interval [0x0, 0xfffff]. 
 Its compressed form, `c.lui`, accepts only those in the subintervals [0x1, 0x1f] and [0xfffe0, 0xfffff]. 
 
-Add Immediate's Immediate
+Signed Immediates for I- and S-Type Instructions
 -------------------------
 
-The immediate argument to `addi` is an integer in the interval [-2048, 2047].
-The integers in the subinterval [-2048, -1] can also be passed by their (unsigned) associates
+All I- and S-type instructions with 12-bit signed immediates
+--- e.g., `addi` but not `slli` ---
+accept their immediate argument as an integer in the interval [-2048, 2047].
+Integers in the subinterval [-2048, -1] can also be passed by their (unsigned) associates
 in the interval [0xfffff800, 0xffffffff] on RV32I, and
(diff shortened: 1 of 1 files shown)


aswaterman:
Suggest "Representation of Signed Immediates for I-type Instructions" or something; as it is, it sounds `addi`-specific.

I thought about generalizing it to N-bit signed immediates, but the extra arithmetic will make it too opaque to be useful.

nick-knight:
I intended it to be `addi`-specific because I didn't verify it exhaustively for all I-type instructions, or prove it by analyzing the binutils source tree. I will certainly take your word for it if this is a general feature.

However, it doesn't appear to work when using the LLVM integrated assembler:
```
$ riscv64-unknown-elf-clang -march=rv32i -integrated-as -c tmp.S
tmp.S:1:14: error: operand must be a symbol with %lo/%pcrel_lo/%tprel_lo modifier or an integer in the range [-2048, 2047]
andi x0, x0, 0xfffff800
```
so perhaps I've stumbled across another GNU vs. LLVM debate.

aswaterman:
All I-type instructions with 12-bit signed immediates.  (e.g. doesn't apply to shift instructions, which have lg(XLEN)-bit unsigned immediates within I-type format)

topperc:
LLVM should match GNU `as` now.

Though I noticed that GNU `as` doesn't support 0xffffffff as a valid constant for c.addi on RV32.

nick-knight:
@topperc Thanks for the update. 

It appears both GNU and LLVM assemblers accept the assembly syntax described in this PR. It's up to the community to decide whether this behavior should be enshrined in our assembly language.

aswaterman (approved):
LGTM.  I'll approve, but I don't want to unilaterally click the merge button.

kito-cheng:
LGTM
@asb @luismarques do you mind that a look if this looks good to you?

luismarques:
LGTM. Shouldn't the commits be squashed during the merge?

luismarques (approved):

nick-knight:
I only have "Read" access to this repo so I can't do anything useful.

kito-cheng:
Gonna to merge, thanks @nick-knight, and thanks @topperc push this forward!

PR #73, opened 2022-01-09
Document .attribute directive

kito-cheng:


riscv-asm.md
```diff
@@ -183,6 +183,40 @@ Directive    | Arguments                      | Description
...
+## <a name=.attribute></a> `.attribute`
+
+`.attribute` directive is used for record information about an object
+file/binary that a linker or runtime loader needs to check compatibility.
+
+More information can refer [attribute section in RISC-V psABI](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-elf.adoc#attributes).
+
+`.attribute` take two argument, first argument of `.attribute` is the symbolic
+name of attribute or the attribute number, the prefix `Tag_RISCV_` can be
+omitted, and second argument can be string or number.
```
asb, line 197: So is `attribute stack_align, "16"` legal? (currently only `attribute stack_align, 16` is accepted in LLVM).

kito-cheng: Oh, good catch, the description too loosely here,  `.attribute stack_align, 16` is the only valid input here, and it will check the attribute number, string value if the tag number is odd and an integer value if the tag number is even.

kito-cheng:
Changes:
- More precisely describe second argument of `.attribute` directive.

kito-cheng:
@luismarques @asb @jrtc27 did you mind did a final round review for this?

riscv-asm.md
```diff
@@ -183,6 +183,41 @@ Directive    | Arguments                      | Description
...
+## <a name=.attribute></a> `.attribute`
+
+`.attribute` directive is used for record information about an object
+file/binary that a linker or runtime loader needs to check compatibility.
+
+More information can refer [attribute section in RISC-V psABI](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-elf.adoc#attributes).
+
+`.attribute` take two argument, first argument of `.attribute` is the symbolic
+name of attribute or the attribute number, the prefix `Tag_RISCV_` can be
+omitted, and second argument is string if attribute number is odd and an integer
```
luismarques, line 197: Why "can be omitted"? Why not just mandate that it's used without the `Tag_RISCV_` prefix? I'm not seeing the value in that flexibility, and if there's no reason we might as well just keep the implementation simpler.

riscv-asm.md
```diff
@@ -183,6 +183,41 @@ Directive    | Arguments                      | Description
 .balign      | b,[pad_val=0]                  | byte align
 .zero        | integer                        | zero bytes
 .variant_cc  | symbol_name                    | annotate the symbol with variant calling convention
+.attribute   | name, value                    | RISC-V object attributes, more detailed description see [.attribute](#.attribute).
+
+## <a name=.attribute></a> `.attribute`
+
+`.attribute` directive is used for record information about an object
```
luismarques, line 190: I'm not a native English speaker myself, so take this with a grain of salt, but I would probably tweak this as follows. Nits:
- "The `.attribute` directive" (add "The")
- "to record" (replace "for" with "to")
- "check for compatibility" (add "for")

riscv-asm.md
```diff
@@ -183,6 +183,41 @@ Directive    | Arguments                      | Description
...
+file/binary that a linker or runtime loader needs to check compatibility.
+
+More information can refer [attribute section in RISC-V psABI](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-elf.adoc#attributes).
```
luismarques, line 193: Nit: "For more information refer to"

riscv-asm.md
```diff
@@ -183,6 +183,41 @@ Directive    | Arguments                      | Description
...
+
+`.attribute` take two argument, first argument of `.attribute` is the symbolic
```
luismarques, line 195: Nits:
- "`.attribute` takes two argument. The first argument" ...
- "The second argument is a string if the attribute number" ...

riscv-asm.md
```diff
@@ -183,6 +183,41 @@ Directive    | Arguments                      | Description
...
+```
+
+Attribute name                | Number | Value type     | Description
+:---------------------------- | :----- | :------------- | :----------
+Tag_RISCV_stack_align         |      4 | uleb128        | Indicates the stack alignment requirement in bytes.
+Tag_RISCV_arch                |      5 | NTBS           | Indicates the target architecture of this object.
+Tag_RISCV_unaligned_access    |      6 | uleb128        | Indicates whether to impose unaligned memory accesses in code generation.
+Tag_RISCV_priv_spec           |      8 | uleb128        | Indicates the major version of the privileged specification.
+Tag_RISCV_priv_spec_minor     |     10 | uleb128        | Indicates the minor version of the privileged specification.
+Tag_RISCV_priv_spec_revision  |     12 | uleb128        | Indicates the revision version of the privileged specification.
```
luismarques, line 220: Maybe the descriptions could link to the appropriate psABI document sections, for more details?

kito-cheng: Good suggestion, I think that could prevent out of sync between psABI spec.

riscv-asm.md
kito-cheng: I would prefer to keep that for compatibility since we already ship that on both GNU toolchain  and LLVM toolchain for a while.

kito-cheng: And I must admit that seems redundant...

kito-cheng:
Changes:
- Apply @luismarques's suggestion.
- Remove attribute table from the doc, just refer to psABI spec to prevent out of sync.

kito-cheng:
Ping, any comments? will merge to trunk 1 week after if no further comment :)

PR #76, opened 2022-02-28
Expand description of registers and their uses

scotws:
Distinguish between the `x`-register view and the ABI used by the programmer; include "RV32E" as reason for strange distribution of register ABI. Minor formatting changes.

### DIFF
--- a/riscv-asm.md
+++ b/riscv-asm.md
@@ -4,9 +4,9 @@
 
 The RISC-V Assembly Programmer's Manual is
 
-  &copy; 2017 Palmer Dabbelt <palmer@dabbelt.com>
-  &copy; 2017 Michael Clark <michaeljclark@mac.com>
-  &copy; 2017 Alex Bradbury <asb@lowrisc.org>
+  &copy; 2017 Palmer Dabbelt <palmer@dabbelt.com>\
+  &copy; 2017 Michael Clark <michaeljclark@mac.com>\
+  &copy; 2017 Alex Bradbury <asb@lowrisc.org>\
 
 It is licensed under the Creative Commons Attribution 4.0 International License
 (CC-BY 4.0). The full license text is available at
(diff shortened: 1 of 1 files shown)


riscv-asm.md
```diff
@@ -26,19 +26,39 @@ and vector registers (V extension).
...
-concerned, by convention certain registers are used for special tasks. In
-assembler, they are given standardized names as part of the RISC-V **application
-binary interface** (ABI). This is what you will usually see in code listings. If
-you really want to see the numeric register names, the `-M` argument to objdump
-will provide them.
+In the first case, the registers are named `x0` to `x31`. All are completely
+equal as general-use registers, except for `x0`, which has a special function:
+Reading it always returns 0 and writes to it are ignored. This allows various
+tricks and simplifications. The program counter `pc` is separate from these
+registers, in contrast to other processors such as the ARM-32.
```
aswaterman, line 37: I realize you didn't introduce the bit about ARM-32, but since you're digging around here, I'd suggest deleting the dependent clause.  We generally shouldn't make reference to other ISAs except when necessary, and it's not necessary here, since ARM is the outlier in this respect, not RISC-V.

scotws: Sure!

kito-cheng:
Generally LGTM, @aswaterman do you mind having a final review for that?

riscv-asm.md
```diff
@@ -26,19 +26,39 @@ and vector registers (V extension).
 
 ## General registers
 
-The RV32I base integer ISA includes 32 registers, named `x0` to `x31`. The
-program counter `PC` is separate from these registers, in contrast to other
-processors such as the ARM-32. The first register, `x0`, has a special function:
-Reading it always returns 0 and writes to it are ignored. As we will see later,
-this allows various tricks and simplifications.
+The **instruction set architecture** (ISA) of RISC-V contains 32 registers.
```
aswaterman, line 29: The old wording was more precise.  It's a true statement that the RV32I and RV64I base ISAs contain 32 registers.  But the other base ISA (RV32E) has only 16 registers, and various ISA extensions (like F) increase the register count beyond 32.

PR #77, opened 2022-04-16
Update assembler relocation function related descriptions

hirooih:
In the table of [A listing of standard RISC-V pseudoinstructions](https://github.com/riscv-non-isa/riscv-asm-manual/blob/master/riscv-asm.md#-a-listing-of-standard-risc-v-pseudoinstructions), for example, the base Instruction for `la rd, symbol` is described as:

```
auipc rd, symbol[31:12]; addi rd, rd, symbol[11:0]
```

In The RISC-V Instruction Set Manual Volume I: Unprivileged ISA, v20191213, Chapter 25 RISC-V Assembly Programmer’s Handbook (in the latest version the chapter is deleted), it is described as;

```
auipc rd, delta[31 : 12] + delta[11]
addi rd, rd, delta[11:0]
where delta = symbol − pc
```

The current description in this manual is wrong. It does not consider the signed immediate offset.

This PR proposes:
- Describe assembler relocation functions more precisely in [Assembler Relocation Functions](https://github.com/riscv-non-isa/riscv-asm-manual/blob/master/riscv-asm.md#assembler-relocation-functions)
- Use assembler relocation functions in the table of  [A listing of standard RISC-V pseudoinstructions](https://github.com/riscv-non-isa/riscv-asm-manual/blob/master/riscv-asm.md#-a-listing-of-standard-risc-v-pseudoinstructions)
- Add an example of negative offset on [Constant](https://github.com/riscv-non-isa/riscv-asm-manual/commit/62250ebce5c8244e3648ac5aac0157165df6caf9)

Note that I could not find clear description of `%tprel_*` and `%tls_*`.  So they maybe still wrong.  Please review them carefully.


johnwinans:
I would prefer to see the LA and LI operations described along the lines of:

li rd,constant

            lui  rd,(constant+0x00000800) >> 12
            addi rd,rd,(constant & 0x00000fff)

Note that the LUI will sign-extend the 20-bit value on RV64 & RV128 systems.

hirooih:
@johnwinans 

Thank you for your suggestion.
I borrowed the expression of the RISC-V specification, but the proposed expression is easier to understand. RISC-V ABIs Specification uses the similar expression.

And I also found I misunderstood the definition of %pcrel_lo(label). I fixed it, too.

hirooih:
> And I also found I misunderstood the definition of %pcrel_lo(label). I fixed it, too.

I did not fix the psedoinstructions table.

PR #83, opened 2023-01-31
The GAS assembler implicitly converts near->far branches

palmer-dabbelt:
This wasn't written down anywhere, but I'm pretty sure we've been doing it since the original port and at least GCC depends on it.

Signed-off-by: Palmer Dabbelt <palmer@rivosinc.com>

### DIFF
--- a/riscv-asm.md
+++ b/riscv-asm.md
@@ -496,6 +496,31 @@ as seen by `objdump`:
    4:	08050513          	addi	a0,a0,128 # 40003080 <UART_BASE>
 ```
 
+Far Branches
+------------
+
+The assembler will convert conditional branches into far branches when
+necessary, via inserting a short branch with inverted conditions past an
+unconditional jump.  For example
+
+```assembly
+target:
+	bne a0, a1, target
+.rep 1024
+	nop
+.endr
+	bne a0, a1, target
+```
+
+ends up as
+
+```
+       0:	00b51063          	bne	a0,a1,0 <target>
+...
+    1004:	00b50463          	beq	a0,a1,100c <target+0x100c>
+    1008:	ff9fe06f          	j	0 <target>
+```
+
 Function Calls
 --------------
 


cmuellner:
I would search for that information in the RISC-V Toolchain Conventions, but since that document references the RISC-V Asm Manual in the Assembler behavior section, this PR is fine.

@kito-cheng, any opinion? If not, I'll merge this PR later this week.

cmuellner (approved):

kito-cheng (commented):
Generally LGTM, but I would like to add some word to mention some compiler e.g. LLVM isn't support that...although the [patch for that](https://reviews.llvm.org/D108961) is restart the discussion recently, but so far all clang/llvm release isn't support that feature.

cmuellner:
Thanks for adding the necessary context!

Here is a summary of the situation:
* GAS does the conversion since day 0
* Upstream LLVM does not do this conversion
* Nobody likes this "magic" but tools depend on it
* Removing the need for this conversion would be much work and break existing code
* Specifying (=requiring) the conversion and implementing it in LLVM would be the easy path
* Having different behavior in GAS and LLVM is what everyone wants to avoid

We can set a deadline until someone can bring up a better solution. But given people repeatedly (every year) stumble into this undocumented and unexpected behavior, we should solve it this time while everyone's attention is raised.

nick-knight:
Is there overlap between @palmer-dabbelt 's PR and this one I filed a few years ago? https://github.com/riscv-non-isa/riscv-asm-manual/pull/58

palmer-dabbelt:
> Is there overlap between @palmer-dabbelt 's PR and this one I filed a few years ago? #58

Looks like it's doing the same thing, just with different words.

kito-cheng:
LLVM support this as well on LLVM 16, so let merge!

Ref:
[1] LLVM patech: https://reviews.llvm.org/D108961

PR #85, opened 2023-04-26
Document fli instruction from zfa extension

kito-cheng:
Split non-zfa part into https://github.com/riscv-non-isa/riscv-asm-manual/pull/86
This is chaining PR with #86

---

Zfa has provide fli instruction to load a floating point immediate, but some of value are hard to written by human, so we would like to support 3 different format of floating value: hexadecimal, decimal, or standard notation formats (a.k.a scientific notation); document this could let user easier know how to use those value in their asm code.

### DIFF
--- a/riscv-asm.md
+++ b/riscv-asm.md
@@ -554,6 +554,23 @@ referenced on [this page](https://en.cppreference.com/w/cpp/language/floating_li
 Load Floating-point Immediate
 -------------------
 
+The `Zfa` extension introduces `fli.{h|s|d|q}` instructions for loading a specific
+set of floating-point immediates, supported values can be found in the
+RISC-V ISA specification but are also listed below.
+
+The `fli` instruction is used to load a floating point immediate into a floating
+register, the accpted immediate is defined in [Floating-point literals](fp-literal),
+and the reference table can be found in [FLI operands reference table](fli-operand-value).
+
+```assembly
+	fli.s	fa0, 0x1p-15
+	fli.s	fa1, 0.00390625
+	fli.s	fa2, 6.25e-02
+```
+
+The tool should reject any value that does not exactly match a floating-point
+immediate operand for the 'fli' instruction.
+
 RISC-V does not offer a generic pseudoinstruction to load an arbitrary floating
 point immediate value. Instead, a programmer can use the `.float`/`.double`
 directive to declare a floating point immediate value in the source code, and
@@ -567,6 +584,48 @@ pseudoinstruction (`fl{h|w|d|q}`).
 	.text
 	flw fa0, .VAL, t0
 ```
+
+
+<a name="fli-operand-value"></a>FLI operands reference table
+
+Value                       | Example legal input values
+:----------------------     | :---------------------
+-1.0                        | -0x1p+0, -1.0, -1.0e+0
+Minimum positive normal     | min
+1.0 x 2 ^ -16               | 0x1p-16, 0.0000152587890625, 1.52587890625e-05
+1.0 x 2 ^ -15               | 0x1p-15, 0.000030517578125, 3.0517578125e-05
+1.0 x 2 ^ -8                | 0x1p-8, 0.00390625, 3.90625e-03
+1.0 x 2 ^ -7                | 0x1p-7, 0.0078125, 7.8125e-03
+0.0625 (2 ^ -4)             | 0x1p-4, 0.0625, 6.25e-02
+0.125 (2 ^ -3)              | 0x1p-3, 0.125, 1.25e-01
+0.25                        | 0x1p-2, 0.25, 2.5e-01
+0.3125                      | 0x1.4p-2, 0.3125, 3.125e-01
+0.375                       | 0x1.8p-2, 0.375, 3.75e-01
+0.4375                      | 0x1.cp-2, 0.4375, 4.375e-01
+0.5                         | 0x1p-1, 0.5, 5.0e-01
+0.625                       | 0x1.4p-1, 0.625, 6.25e-01
+0.75                        | 0x1.8p-1, 0.75, 7.5e-01
+0.875                       | 0x1.cp-1, 0.875, 8.75e-01
+1.0                         | 0x1p+0, 1.0, 1.0e+00
+1.25                        | 0x1.4p+0, 1.25, 1.25e+00
+1.5                         | 0x1.8p+0, 1.5, 1.5e+00
+1.75                        | 0x1.cp+0, 1.75, 1.75e+00
+2.0                         | 0x1p+1, 2.0, 2.0e+00
+2.5                         | 0x1.4p+1, 2.5, 2.5e+00
+3                           | 0x1.8p+1, 3.0, 3.0e+00
+4                           | 0x1p+2, 4.0, 4.0e+00
+8                           | 0x1p+3, 8.0, 8.0e+00
+16                          | 0x1p+4, 16.0, 1.6e+01
+128 (2 ^ 7)                 | 0x1p+7, 128.0, 1.28e+02
+256 (2 ^ 8)                 | 0x1p+8, 256.0, 2.56e+02
+2 ^ 15                      | 0x1p+15, 32768.0, 3.2768e+04
+2 ^ 16                      | 0x1p+16, 65536.0, 6.5536e+04
+Positive infinity           | inf
+Canonical NaN               | nan
+
+A value can be expressed in various forms within the same format. For example,
+6.5536e+04 can be alternatively written as 6553.6e+01 or 65.536e+03. The table
+provides one possible representation, but any equivalent exact value may be used.
 
 Load Address
 ------------


kito-cheng:
@asb @cmuellner @topperc @a4lg

riscv-asm.md
```diff
@@ -525,6 +528,82 @@ Integers in the subinterval [-2048, -1] can also be passed by their (unsigned) a
...
+```
+
+RISC-V does not offer a generic pseudoinstruction to load an arbitrary floating
+point immediate value. Instead, a programmer can use the `.float`/`.double`
+directive to declare a floating point immediate value in the source code, and
+then load it into a floating point register using the load global
+pseudoinstruction (`fl{h|w|d|q}`).
+
+RISC-V didn't provide generic pseudoinstruction to load arbitrary floating point
+immediate value, programmer can use `.float`/`.double` diretive to declare a
```
cmuellner, line 593: diretive -> directive

riscv-asm.md
```diff
@@ -525,6 +528,82 @@ Integers in the subinterval [-2048, -1] can also be passed by their (unsigned) a
...
+float point immediate within the source, and then load into floating point
```
cmuellner, line 594: float -> floating

cmuellner: and the load into -> and then load it into a

riscv-asm.md
```diff
@@ -525,6 +528,82 @@ Integers in the subinterval [-2048, -1] can also be passed by their (unsigned) a
...
+
+```assembly
+	.data
+.VAL:
+	.float .0x1p+17
+	.text
+	flh fa0, .VAL, t0
+```
+
+NOTE: `Zfa` is NOT ratified yet.
```
cmuellner, line 605: I would remove this comment (and optionally wait for Zfa to be ratified).

riscv-asm.md
```diff
@@ -525,6 +528,82 @@ Integers in the subinterval [-2048, -1] can also be passed by their (unsigned) a
...
+The `fli` instruction is used to load a floating point immediate into a floating
+register. The floating point immediate must be one of the values listed in the
+above table and can be expressed in hexadecimal, decimal, or standard notation
+(also known as scientific notation).
+
+```assembly
+	fli.s	fa0, 0x1p-15
+	fli.s	fa1, 0.00390625
+	fli.s	fa2, 6.250000e-02
+```
```
cmuellner, line 584: What about `62.5e-3`?

riscv-asm.md
```diff
@@ -525,6 +528,82 @@ Integers in the subinterval [-2048, -1] can also be passed by their (unsigned) a
 in the interval [0xfffff800, 0xffffffff] on RV32I, and
 in [0xfffffffffffff800, 0xffffffffffffffff] on both RV32I and RV64I.
 
+Load Floating Point Immediate
+-------------------
+
+`Zfa` extension has provide `fli.{h|s|d|q}` instruction for loading a specific
+set of floating-point immediates.
+
+The supported floating point immediates are listed in the following table:
```
cmuellner, line 537: Do we need to duplicate the list here (i.e. can we not refer to the specification)?

riscv-asm.md
```diff
@@ -525,6 +528,82 @@ Integers in the subinterval [-2048, -1] can also be passed by their (unsigned) a
 in the interval [0xfffff800, 0xffffffff] on RV32I, and
 in [0xfffffffffffff800, 0xffffffffffffffff] on both RV32I and RV64I.
 
+Load Floating Point Immediate
+-------------------
+
+`Zfa` extension has provide `fli.{h|s|d|q}` instruction for loading a specific
```
cmuellner, line 534: Does this need to be Zfa-specific?

We could instead define a generic way of supported floating-point literal formats supported by the assembler.
Example uses of such floating-point literals would be in `.single <expr>` and `fli.s fN, <expr>`.

The restriction of supported values for `fli` is a restriction that is instruction specific and does not need to be documented here.

cmuellner (commented):
Thanks for the proposal.
I think it is a good idea to document the supported floating-point literal formats that are supported by the assembler.

riscv-asm.md
kito-cheng: Damm, I just forgot remove the original version, plz just ignore this paragraph.

riscv-asm.md
kito-cheng: Same here

riscv-asm.md
kito-cheng: Yeah, I thought same thing, that will require one extra temporal register to hold the address, but that should not really big deal.

riscv-asm.md
kito-cheng: I would like to hold the this here to let user has a way to quick look up how to write those value in any format (he like), since ISA spec isn't provide this table.

riscv-asm.md
kito-cheng: Hmm, sounds like good idea and I could split non-zfa part into a separated PR.

riscv-asm.md
kito-cheng: Tried on both LLVM and binutils, they accept that as well, but I guess we don't need to document that here since it's kind of standard way?

cmuellner: With this question, I wanted to show that an enumeration of valid values is not optimal.

Instead we could reference the floating-point literal formats that are supported. A good overview can be found here: https://en.cppreference.com/w/cpp/language/floating_literal.

riscv-asm.md
```diff
@@ -525,6 +528,82 @@ Integers in the subinterval [-2048, -1] can also be passed by their (unsigned) a
...
+2 ^ 15                      | 0x1p+15                        | 32768.0               | 3.276800e+04
+2 ^ 16                      | 0x1p+16                        | 65536.0               | 6.553600e+04
+Positive infinity           | inf                            | inf                   | inf
+Canonical NaN               | nan                            | nan                   | nan
+
+
+The `fli` instruction is used to load a floating point immediate into a floating
+register. The floating point immediate must be one of the values listed in the
+above table and can be expressed in hexadecimal, decimal, or standard notation
+(also known as scientific notation).
```
cmuellner, line 578: Is your "decimal" representation not a special case of the "standard" notation?

When looking into `man strtof` I only see a decimal format and a hexadecimal format.
E.g. also `-1.` and `-0x1.` are valid floating point literals, but would not fit any format of your table.

riscv-asm.md
```diff
@@ -525,6 +528,82 @@ Integers in the subinterval [-2048, -1] can also be passed by their (unsigned) a
...
+`Zfa` extension has provide `fli.{h|s|d|q}` instruction for loading a specific
+set of floating-point immediates.
+
+The supported floating point immediates are listed in the following table:
+
+Value                       | Hexadecimal                    | Decimal               | Standard Representation / Scientific Notation
+:----------------------     | :---------------               | :-------------------  |
+-1.0                        |-0x1p+0                         | -1.0                  | -1.000000e+00
+Minimum positive normal     | min                            | min                   | min
+1.0 x 2 ^ -16               | 0x1p-16                        | 0.0000152587890625    | 1.525879e-05
```
aswaterman, line 543: Re: `1.525879e-05`: I thought we had agreed to disallow inexact constants, so that this variant would be rejected but `1.52587890625e-06` would be accepted?

kito-cheng:
Split non-zfa part into #86

riscv-asm.md
kito-cheng: My bad, this table is generated from my C code, I did check output for decimal but not well check on Scientific Notation form

aswaterman: Cool.

riscv-asm.md
```diff
@@ -525,6 +528,82 @@ Integers in the subinterval [-2048, -1] can also be passed by their (unsigned) a
...
+immediate value, programmer can use `.float`/`.double` diretive to declare a
+float point immediate within the source, and then load into floating point
+register by load global pseudoinstruction (`fl{h|w|d|q}`) to load the value.
+
+```assembly
+	.data
+.VAL:
+	.float .0x1p+17
+	.text
+	flh fa0, .VAL, t0
```
topperc, line 602: is flh correct for loading from .float?

kito-cheng: thanks, that should be flw

riscv-asm.md
kito-cheng: Added a note, this should improved:
```
NOTE: A value can be expressed in different forms within the same format.
For example, 6.5536e+04 can be written as 6553.6e+01 or 65.536e+03. The table
lists one possible representation, but any exact value can be used.
```

riscv-asm.md
kito-cheng: reoreded the paragraph, reference to RISC-V ISA spec instead, but put the table in the button of this section and mark as NOTE.

kito-cheng:
Changes:
- Split non-zfa part into https://github.com/riscv-non-isa/riscv-asm-manual/pull/86
- Add nominate text to say tool should reject inexact value.
- Add note to say a value written in any form that has any exact value (e.g. 6.5536e+04 can be written as 6553.6e+01 or 65.536e+03)
- Add https://en.cppreference.com/w/cpp/language/float this as ref.

a4lg:
Thanks for proposal @kito-cheng!

It's good to hear that the constant operand must be exact (if not, I would be worried about its precision for fuzzy matching).

kito-cheng:
Changes:
- Reference to `Floating-point literals` instead of reference to something like hexadecimal.
- Revise the wording.

asb:
I think it's helpful to have this documented somewhere, but defer to others on detailed line-by-line correctness review.

riscv-asm.md
```diff
@@ -545,6 +545,23 @@ referenced on [this page](https://en.cppreference.com/w/cpp/language/floating_li
 Load Floating-point Immediate
 -------------------
 
+The `Zfa` extension introduces `fli.{h|s|d|q}` instruction for loading a specific
```
asb, line 548: instruction => instructions

riscv-asm.md
```diff
@@ -545,6 +545,23 @@ referenced on [this page](https://en.cppreference.com/w/cpp/language/floating_li
...
+set of floating-point immediates, supported value can be found from the RISC-V
```
asb, line 549: value => values. Though we list all the values below, so perhaps "supported values can be found in the RISC-V ISA specification but are also listed below."

kito-cheng:
@asb thanks, suggestion applied :)

kito-cheng:
ping :)

riscv-asm.md
```diff
@@ -568,6 +585,48 @@ pseudoinstruction (`fl{h|w|d|q}`).
 	flw fa0, .VAL, t0
 ```
 
+
+<a name="fli-operand-value"></a>FLI operands reference table
+
+Value                       | Possible inputs values
```
aswaterman, line 591: I suggest changing "Possible inputs values" to "Example legal input values".  This makes it clear these aren't the only possibilities.  (Also note that "input" should not be plural.)

aswaterman (approved):
LGTM modulo my last suggestion.

cmuellner (approved):
LGTM (after applying the change requested by @aswaterman).

kito-cheng:
Updated, thanks :)

PR #86, opened 2023-04-27
Document floating point directive and generic way to to load a floating point immediate to floating point register

kito-cheng:
Split from #85, this PR only contain non-zfa part.

Contain two things:
1. `.float` and `.double`
2. Generic way to to load a floating point immediate to floating point register via load from memeory.

### DIFF
--- a/riscv-asm.md
+++ b/riscv-asm.md
@@ -183,8 +183,8 @@ Directive    | Arguments                      | Description
 .8byte       | expression [, expression]*     | 64-bit comma separated words
 .dword       | expression [, expression]*     | 64-bit comma separated words
 .quad        | expression [, expression]*     | 64-bit comma separated words
-.float       | expression [, expression]*     | 32-bit floating point values
-.double      | expression [, expression]*     | 64-bit floating point values
+.float       | expression [, expression]*     | 32-bit floating point values, see [Floating-point literals](#fp-literal) for the value format.
+.double      | expression [, expression]*     | 64-bit floating point values, see [Floating-point literals](#fp-literal) for the value format.
 .dtprelword  | expression [, expression]*     | 32-bit thread local word
 .dtpreldword | expression [, expression]*     | 64-bit thread local word
 .sleb128     | expression                     | signed little endian base 128, DWARF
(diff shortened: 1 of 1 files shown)


kito-cheng:
ping @asb @cmuellner  :)

kito-cheng:
Changes:
- Add new section: Floating-point literals

asb:
I think we risk veering into becoming a general assembly guide which might be more of a burden than we want to take on, but documenting .float and .double seems handy. I wonder a bit about "The Assembler supports the same floating point literal formats as those defined in the C and C++ standards" - find for gas and LLVM as and any other assembler wanting to consume the same input, but not necessarily true of any arbitrary assembler. Perhaps this document needs a "Scope" section that says something like "This document aims to provide guidance to assembly programmers targeting common open source RISC-V assemblers such as GNU as and LLVM's assembler. Other assemblers might not support the same directives or pseudoinstructions."

kito-cheng:
Thanks, although one of my ultimate goal is make this document become RISC-V standard Assembly manual, but archive that apparently need much more effort than I/we can pay, so limit the scope for this document is sound good idea, let me create PR in my next workday :)

kito-cheng:
Add #89, @asb did you have any further concern about this PR?

asb:
With #89 in place this is looking pretty good to me, but given you only document `.float` and `.double` perhaps it doesn't make sense to list `flh` and `flq` (or to instead give guidance on using .2byte/.8byte or similar with the bit pattern of the encoded FP value.

riscv-asm.md
```diff
@@ -525,6 +527,38 @@ Integers in the subinterval [-2048, -1] can also be passed by their (unsigned) a
...
+referenced on [this page](https://en.cppreference.com/w/cpp/language/floating_literal)
+
+Load Floating-point Immediate
+-------------------
+
+RISC-V does not offer a generic pseudoinstruction to load an arbitrary floating
+point immediate value. Instead, a programmer can use the `.float`/`.double`
+directive to declare a floating point immediate value in the source code, and
+then load it into a floating point register using the load global
+pseudoinstruction (`fl{h|s|d|q}`).
```
aswaterman, line 552: `flw` not `fls`, right?

riscv-asm.md
```diff
@@ -525,6 +527,38 @@ Integers in the subinterval [-2048, -1] can also be passed by their (unsigned) a
...
+
+```assembly
+	.data
+.VAL:
+	.float .0x1p+17
+	.text
+	fls fa0, .VAL, t0
```
aswaterman, line 559: ditto

riscv-asm.md
nick-knight, line 552: Same comment as @asb : you only mention the `.float` and `.double` directives; what should I use in the FP16 and FP128 cases? Note that `.quad` is already used by GAS. Maybe `.float{16,32,64,128}` would be a more scalable approach?

kito-cheng: Oh, the reason I document `.float` and `.double` here is because `.float16` is not supported by clang, and `.float128` are not supported for both clang and binutils...

`.quad` is supported by both, I think I should add that to doc.

kito-cheng:
Changes:
- Fix fls -> flw
- Add `.quad` for 128 bit floating point

aswaterman (approved):

riscv-asm.md
kito-cheng: My thought is we should add `.float{16,32,64,128}`  and `.bfloat16` in future, but anyway it out of scope of this PR, let me finish fli stuff first.

kito-cheng:
@aswaterman thanks bro :)

PR #87, opened 2023-04-27
Update flw to fls

kito-cheng:
flw is kind of legacy mnemonic, use fls instead.


cmuellner:
The ISA spec still names this instructions FLW and FSW (see https://github.com/riscv/riscv-isa-manual/blob/master/src/f.tex#L327).

Why would we not follow the spec?

kito-cheng:
ohhh, my mistake, fls should be the deprecated one, not flw.

kito-cheng:
I guess I confused because fli use s rather than w.

PR #92, opened 2023-10-09
Fix grammar error

visitorckw:
Changed 'got' to 'get' for proper grammar.


cmuellner:
I'm afraid that this change is not sufficient to address all grammar mistakes in this sentence.
Probably this would be a better alternative:
> Another example is `.option arch, rv64ifd` + `.option arch, -f`, which results in `rv64ifd`, because `f` will be added back when adding the implied extensions of `d`.

visitorckw:
Thank you for your review and suggestions!
I have made the changes accordingly.

cmuellner (approved):
LGTM

cmuellner:
Thanks for the fix!

PR #93, opened 2024-01-16
Change scratch register of tail to t2 when Zicfilp enabled.

yetingk:
This change is to make `tail` conform with software guarded jump of [Zicfilp]. The reason to not choose `t1` as the label register is that `t1` is also as `.got.plt` offset of `_dl_runtime_resolve` in [PLT].

[Zicfilp]: https://github.com/riscv/riscv-cfi/blob/main/cfi_forward.adoc
[PLT]: https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-elf.adoc#procedure-linkage-table

### DIFF
--- a/riscv-asm.md
+++ b/riscv-asm.md
@@ -802,7 +802,7 @@ the current position:
 
 [^3]: `ra` is implicitly used to save the return address.
 [^4]: similar to `call <symbol>`, but `<rd>` is used to save the return address instead.
-[^5]: `t1` is implicitly used as a scratch register.
+[^5]: If the `Zicfilp` extension is available, `t2` is implicitly used as a scratch register. Otherwise,`t1` is implicitly used as a scratch register.
 [^6]: similar to `tail <symbol>`, but `<rt>` is used as the scratch register instead.
 
 The following example shows how these pseudoinstructions are used:
@@ -964,7 +964,7 @@ jr rs                        | jalr x0, rs, 0                                   
 jalr rs                      | jalr x1, rs, 0                                                | Jump and link register
 ret                          | jalr x0, x1, 0                                                | Return from subroutine
 call offset                  | auipc x1, offset[31:12] <br>jalr x1, x1, offset[11:0]         | Call far-away subroutine
-tail offset                  | auipc x6, offset[31:12] <br>jalr x0, x6, offset[11:0]         | Tail call far-away subroutine
+tail offset                  | auipc x6, offset[31:12] <br>jalr x0, x6, offset[11:0]         | Tail call far-away subroutine | It will use `x7` as scratch register when `Zicfilp` extension is available.
 fence                        | fence iorw, iorw                                              | Fence on all memory and I/O
 pause                        | fence w, 0                                                    | PAUSE hint
 


riscv-asm.md
```diff
@@ -793,7 +793,7 @@ the current position:
 
 [^3]: `ra` is implicitly used to save the return address.
 [^4]: similar to `call <symbol>`, but `<rd>` is used to save the return address instead.
-[^5]: `t1` is implicitly used as a scratch register.
+[^5]: `t2` is implicitly used as a scratch register.
```
kito-cheng, line 796: Don't break backward compatibility, try to add rather than just change:
"`t1` is implicitly used as a scratch register, but if the Zicfiss extension is enabled, it will use t2 instead."

aswaterman (approved):
OK with me, but I'd like to know if there actually is a problem with unconditionally changing to `t2`.  Having both options seems like an unnecessary complication (although I concede it's only a very minor complication).

kito-cheng:
cc @ved-rivos @jrtc27 @asb

kito-cheng:
@aswaterman The reason I suggest keep both is I don't want make older tool become non-conformance with this asm manual, but maybe we could using different way/wording to prevent that?

jrtc27:
I still believe the extension should have been defined in a way that doesn't break existing practices, and it's pretty weird to have `tail` change which register it uses based on what extensions you happen to have enabled. That sounds like asking for trouble if people are assuming the register that `tail` clobbers (thinking about custom calling conventions when branching to a non-preemptible symbol), too, given it has been explicitly specified as being t1...

jrtc27:
Without thinking through all the implications, my instincts are that it is much better to change the PLT implementation (which is technically also specified today, but really shouldn't be, the code should just be an example of a possible implementation, as application code shouldn't know/care about it) than to change the semantics of pseudo-instructions.

nick-knight (commented):
Whatever change is made here should also be reflected in [the table](https://github.com/riscv-non-isa/riscv-asm-manual/blob/master/riscv-asm.md#-a-listing-of-standard-risc-v-pseudoinstructions).

ved-rivos:
Looks good to me.

kito-cheng (approved):
Ideally we should keep those pseudo instructions consistency in every situation, but I know it's kinda trade off in ISA design, so LGTM from my side as well.

riscv-asm.md
```diff
@@ -793,7 +793,7 @@ the current position:
 
 [^3]: `ra` is implicitly used to save the return address.
 [^4]: similar to `call <symbol>`, but `<rd>` is used to save the return address instead.
-[^5]: `t1` is implicitly used as a scratch register.
+[^5]: `t1` is implicitly used as a scratch register, but if the `Zicfilp` extension is enabled, it will use `t2` instead.
```
topperc, line 796: I don't think I like this sentence, but I'm struggling to make it better.

Maybe

`If the `Zicfilp` extension is enabled, `t2` is implicitly used as a scratch register. Otherwise, `t1` is implicitly used as a scratch register.`

riscv-asm.md
```diff
@@ -954,7 +954,7 @@ jr rs                        | jalr x0, rs, 0
 jalr rs                      | jalr x1, rs, 0                                                | Jump and link register
 ret                          | jalr x0, x1, 0                                                | Return from subroutine
 call offset                  | auipc x1, offset[31:12]; jalr x1, x1, offset[11:0]            | Call far-away subroutine
-tail offset                  | auipc x6, offset[31:12]; jalr x0, x6, offset[11:0]            | Tail call far-away subroutine
+tail offset                  | auipc x6, offset[31:12]; jalr x0, x6, offset[11:0]            | Tail call far-away subroutine | It will use `x7` as scratch register when `Zicfilp` extension enabled.
```
topperc, line 957: "when `Zicfilp` extension enabled" -> "when the `Zicfilp` extension is available"

kito-cheng:
NOTE: public review period for zicfiss and zicfilp are ended, waiting for ratification flow and then we will merge this.

cmuellner:
Zicfiss, Zicfilp were ratified in June.

PR #99, opened 2024-04-18
Add %lpad_hash for Zicfilp

kito-cheng:
NOTE: This PR will keep in draft state until toolchain PoC and psABI spec ready.

---

Zicfilp has provided two labeling schemes: simple and complex (also known as function signature-based). The simple scheme uses an lpad with a constant 0, which does not require any hashing mechanism. In contrast, the complex labeling scheme computes the MD5 hash from the signature string.

Filling up an MD5 hash value is straightforward for compilers, but it is non-trivial work for humans to maintain. Therefore, we have added new assembler modifiers to compute this value.

See also: https://github.com/riscv/riscv-cfi/issues/151

### DIFF
--- a/riscv-asm.md
+++ b/riscv-asm.md
@@ -397,9 +397,10 @@ linker relaxation accidentally if user already disable linker relaxation.
 
 Push/pop current options to/from the options stack.
 
-## Assembler Relocation Functions
+## Assembler Modifiers
 
-The following table lists assembler relocation expansions:
+The RISC-V assembler supports the following modifiers for relocatable addresses
+or for computing special values used in RISC-V instruction operands:
 
 Assembler Notation          | Description                    | Instruction / Macro
 :----------------------     | :---------------               | :-------------------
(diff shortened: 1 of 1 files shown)


riscv-asm.md
```diff
@@ -397,9 +397,10 @@ linker relaxation accidentally if user already disable linker relaxation.
 
 Push/pop current options to/from the options stack.
 
-## Assembler Relocation Functions
+## Assembler Modifiers
```
kito-cheng, line 400: This term is learned from GNU as manual https://sourceware.org/binutils/docs/as/RISC_002dV_002dModifiers.html

mylai-mtk:
Actually, I wonder if this would turn out to be useful, since the function signature string format is (for now) drafted to be the mangled string of the function type, which IMHO is not human-friendly enough to be written/read correctly by a human programmer. Also, AFAIK, there's no convenient tool to mangle function signatures to strings, despite the widespread usage of `c++filt` for demangling. Under this impression, I doubt there would be people trying use this assembly modifier due to the difficulty of producing those function signature strings. 

To get a feel: `lpad %lpad_hash("FiiPPcE")` vs `lpad 0xe088e`. While the `%lpad_hash()` form may be easier to read for someone familiar with the mangling rule (this advantage may be useful in the rare scenario of reviewing assemblies by compiler experts), I don't think it's comprehensible for average assembly developers, not to mention to write it out without the help of tools.

Though I doubt the usefulness of this `%lpad_hash()` modifier, I do agree that we need a method to ease the pain of obtaining correct label values. Here I propose a possible, but not really ideal method: In my own toolchain prototyping process, I emit symbols containing the label values for all C/C++ functions, so I can easily extract the resulting compiler-generated labels by looking into symbol tables. This emission was originally intended to facilitate function-signature-based PLT generation in linkers, but it turns out that I use it a lot to when patching musl libc assemblies with `lpad` insns. This "compile-then-inspect-binary" approach is far from straightforward and beautiful, but at least I can trust the values obtained to be correct, if I don't make a mistake when copying them 😜

ved-rivos:
> While the %lpad_hash() form may be easier to read for someone familiar with the mangling rule (this advantage may be useful in the rare scenario of reviewing assemblies by compiler experts), I don't think it's comprehensible for average assembly developers, not to mention to write it out without the help of tools.

Should we add a modifier that takes the function prototype string instead of mangled string as input - like  %lpad_hash("void (*f)(int, char)").

mylai-mtk:
> Should we add a modifier that takes the function prototype string instead of mangled string as input - like %lpad_hash_proto("void (*f)(int, char)").

I guess this would be a huge effort for assemblers to implement, since they do not know anything about the C/C++ language, so the corresponding C parser and C++ mangler would need to be pulled in, which is a big cost for a minor convenient feature like this.

riscv-asm.md
```diff
@@ -410,11 +411,15 @@ Assembler Notation          | Description                    | Instruction / Mac
...
-\* These reuse %pcrel_lo(label) for their lower half
+%tls_ie_pcrel_hi(symbol) \*1 | TLS IE "Initial Exec" (HI20)   | auipc
+%tls_gd_pcrel_hi(symbol) \*1 | TLS GD "Global Dynamic" (HI20) | auipc
+%got_pcrel_hi(symbol) \*1    | GOT PC-relative (HI20)         | auipc
+%lpad_hash("func_signature") \*2 | Hash for landing pad labeling | lpad
+
+\*1 These reuse %pcrel_lo(label) for their lower half
+\*2 The format of the function signature string is defined in the RISC-V psABI.
+    The hash result is obtained from the lower 20 bits of the MD5 result of the
+    string.
```
mylai-mtk, line 422: Technically speaking, this is not a complete description of the hashing algorithm, so I suggest you either explain more, or remove this message

PR #100, opened 2024-05-01
Fix references to B extension.

topperc:
These should reference Zba and Zbb intead.


aswaterman (approved):
Thanks for the improvement!

aswaterman:
Looks like my write permission to this repo has disappeared, so someone else will need to approve and merge.

cmuellner (approved):
LGTM

Issue #98, opened 2024-04-04
Two suggested improvements to the pseudo instruction table


secworks:
I would like to suggest two improvements to the table with pseudo instructions:

1. For pseudo instructions that maps to more than one real instruction, put the real instructions on separate lines. Today the real instructions are separated with ';', and the second instruction is wrapped half way onto the next line. This makes it quite hard to see the real instructions and their operands, fields.

2. Somewhere, either before or after the table explain 'myriad sequences', an expression which only appears in the table. For non-native English speakers, this short expression may be hard to understand what it implies.

cmuellner:
Both suggestions sound good.
Are you interested in providing PRs as well?

cmuellner:
Fixed with #101.

secworks:
Looks really good!

PR #101, opened 2024-05-09
Fixed issue #98

apazos:
Fixed the two issues reported in issue #98:

1. For pseudo instructions that map to more than one real instruction, placed the real instructions on separate lines
2. Added a footnote about 'myriad sequences' expression


cmuellner (approved):
LGTM and also renders well in the browser.

kadiwa4:
The footnote doesn't work :/

You forgot the colon after the definition of `[^1]`, and also you should choose a different number because `[^1]` is already defined elsewhere

apazos:
Thanks for testing and pointing it out, @kadiwa4 , will push a fix.

PR #105, opened 2024-06-26
Recommend avoiding .align

Timmmm:
Move the `.p2align` and `.balign` directives next to `.align` in the table, and add a note recommending avoiding `.align` in favour of the explicit directives which are much clearer.

### DIFF
--- a/riscv-asm.md
+++ b/riscv-asm.md
@@ -153,7 +153,9 @@ The following table lists assembler directives:
 
 Directive    | Arguments                      | Description
 :----------- | :-------------                 | :---------------
-.align       | integer                        | align to power of 2 (alias for .p2align)
+.align       | integer                        | align to power of 2 (alias for `.p2align` which is preferred - see [.align](#.align))
+.p2align     | p2,[pad_val=0],max             | align to power of 2
+.balign      | b,[pad_val=0]                  | byte align
 .file        | "filename"                     | emit filename FILE LOCAL symbol table
 .globl       | symbol_name                    | emit symbol_name to symbol table (scope GLOBAL)
 .local       | symbol_name                    | emit symbol_name to symbol table (scope LOCAL)
@@ -190,11 +192,17 @@ Directive    | Arguments                      | Description
 .dtpreldword | expression [, expression]*     | 64-bit thread local word
 .sleb128     | expression                     | signed little endian base 128, DWARF
 .uleb128     | expression                     | unsigned little endian base 128, DWARF
-.p2align     | p2,[pad_val=0],max             | align to power of 2
-.balign      | b,[pad_val=0]                  | byte align
 .zero        | integer                        | zero bytes
 .variant_cc  | symbol_name                    | annotate the symbol with variant calling convention
 .attribute   | name, value                    | RISC-V object attributes, more detailed description see [.attribute](#.attribute).
+
+## <a name=.align></a> `.align`
+
+The `.align` directive for RISC-V is an alias to `.p2align`, which aligns to a
+power of two, so `.align 2` means align to 4 bytes. Because the definition of
+the `.align` directive [varies by architecture](https://sourceware.org/binutils/docs/as/Align.html),
+it is recommended to use the unambiguous `.p2align` or `.balign` directives
+instead.
 
 ## <a name=.attribute></a> `.attribute`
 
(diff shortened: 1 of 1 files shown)


cmuellner (approved):
I think that's a good idea.

@kito-cheng any objections?

aswaterman (changes requested):
Please remove the editorialization about `.align`.  `.align` is standard stuff in GAS etc.  I'm not unsympathetic to your point, but it's more important that RISC-V be compatible with other assembly languages than it is to scratch this itch.

Timmmm:
> Please remove the editorialization about `.align`. `.align` is standard stuff in GAS etc. I'm not unsympathetic to your point, but it's more important that RISC-V be compatible with other assembly languages than it is to scratch this itch.

It is still compatible with other assembly languages, at least in the sense that you can write `.align`. It's not compatible in the sense that it won't necessarily do the same thing though, which was the whole point of the comment. This is only a recommendation anyway, so I don't see how it could affect compatibility. Maybe I've missed your point.

cmuellner:
> Please remove the editorialization about `.align`. `.align` is standard stuff in GAS etc. I'm not unsympathetic to your point, but it's more important that RISC-V be compatible with other assembly languages than it is to scratch this itch.

The whole PR is just an editorialization about `.align`.
Moving `.p2align` and `.balign` were already documented.
So, dropping the editorialization means dropping the PR.

The [GAS manual](https://sourceware.org/binutils/docs/as/Align.html) states:
> The way the required alignment is specified varies from system to system.

So, there can't be compatibility with all other supported architectures/systems.

But I think I see what you mean. In fact, RISC-V defines `.align` in a reasonable way, and it did not cause the inconsistent ways in which `.align` was defined for other architectures in the past. Therefore, it is reasonable to question why we would recommend not using `.align`.

Timmmm:
> Therefore, it is reasonable to question why we would recommend not using .align.

Clearly my explanation wasn't as good as I thought! It's because the meaning of `.align` is not clear without carefully reading the manual, and this can lead to bugs and misunderstandings (not theoretical; I have seen this). I think many people would assume `.align 4` means align to 4 bytes (which is not unreasonable given that it *does* mean that on other architectures!).

It technically doesn't make any difference to RISC-V how other architectures define `.align`, but the fact that they are inconsistent adds to the ambiguity.

I'm totally open to better ways to explain that if anyone can think of any. I was trying to keep it a bit briefer than this comment...

riscv-asm.md
```diff
@@ -190,12 +192,21 @@ Directive    | Arguments                      | Description
...
 .uleb128     | expression                     | unsigned little endian base 128, DWARF
-.p2align     | p2,[pad_val=0],max             | align to power of 2
-.balign      | b,[pad_val=0]                  | byte align
 .zero        | integer                        | zero bytes
 .variant_cc  | symbol_name                    | annotate the symbol with variant calling convention
 .attribute   | name, value                    | RISC-V object attributes, more detailed description see [.attribute](#.attribute).
 
+## <a name=.align></a> `.align`
+
+The `.align` directive for RISC-V is an alias to `.p2align` which aligns to
```
MaskRay, line 201: Perhaps this can be simplified to:

The interpretation of `.align` [depends on the architecture] .... The directive for RISC-V means ....

The verbose "This can be very confusing"  isn't necessary.

aswaterman:
I meant editorialization in the sense that it reads like an opinion piece rather than a technical document :)  I'd be fine with something like the following, along the lines of what @MaskRay was suggesting:


The `.align` directive for RISC-V is an alias to `.p2align`, which aligns to a power of two, so `.align 2` means align to 4 bytes.  Because the definition of the `.align` directive varies by architecture, it is recommended to use the unambiguous `.p2align` or `.balign` directives instead.

Timmmm:
Updated, thanks for the suggestion!

aswaterman (approved):
👍

cmuellner (approved):
LGTM

The PR has an independent whitespace change at the end.
If you care, feel free to fix that (e.g. splitting that whitespace change off into its own commit).
Otherwise, I'll merge as is in a couple of hours.

MaskRay:
Typo in the title: recommend

MaskRay (approved):

riscv-asm.md
```diff
@@ -190,12 +192,18 @@ Directive    | Arguments                      | Description
...
 .uleb128     | expression                     | unsigned little endian base 128, DWARF
-.p2align     | p2,[pad_val=0],max             | align to power of 2
-.balign      | b,[pad_val=0]                  | byte align
 .zero        | integer                        | zero bytes
 .variant_cc  | symbol_name                    | annotate the symbol with variant calling convention
 .attribute   | name, value                    | RISC-V object attributes, more detailed description see [.attribute](#.attribute).
 
+## <a name=.align></a> `.align`
+
+The `.align` directive for RISC-V is an alias to .p2align, which aligns to a
```
MaskRay, line 201: backticks around `.p2align`

Timmmm: Fixed, thanks

kito-cheng (approved):
LGTM :)

cmuellner:
Merged. Thanks!

Mudassir10X (approved):

PR #111, opened 2024-10-25
Add V extension pseudoinstructions

### LINKED ISSUE #110: Missing pseudo-instructions?
I've been digging for pseudoinstructions in [binutils ](https://github.com/bminor/binutils-gdb/blob/master/opcodes/riscv-opc.c )and found that several pseudoinstructions implemented there are not found in the ASM Manual, is this on purpose?

For reference, these are some, but not all of those pseudo-instructions:

bs.v
vfneg.v
vmclr.m
vmfge.vv 
vmfgt.vv 
vmmv.m 
vmnot.m 
vmset.m
vmsge.vi
vmsge.vv
vmsge.vx
vmsgeu.vi
vmsgeu.vv
vmsgeu.vx
vmsgt.vv
vmsgtu.vv 
vmslt.vi
vmsltu.vi
vneg.v
vnot.v
vt.maskc
vt.maskcn
vwcvt.x.x.v
vwcvtu.x.x.v

AFOliveira:
Closes #110 .

Preview: [riscv-asm.pdf](https://github.com/user-attachments/files/17522971/riscv-asm.pdf)

### DIFF
--- a/src/asm-manual.adoc
+++ b/src/asm-manual.adoc
@@ -1105,17 +1105,11 @@ srli rd, rd, XLEN - 32
 |vl4r.v v4,x0                 | vl4re8.v v4, x0                                               | Equal to vl4re8.v |
 |vl8r.v v8,x0                 | vl8re8.v v8, x0                                               | Equal to vl8re8.v |
 
-|vmsge{u}.vx vd, va, x|
- addi t0, x, -1 + 
- vmsgt{u}.vx vd, va, t0, vm  
-| Vector >= scalar
-| When  x > minimum
-
 |vmsge{u}.vx vd, va, x        
 |vmslt{u}.vx vd, va, x +
 vmnand.mm vd, vd, vd
 | Vector >= scalar, unmasked
-| For any x
+| 
 
 | vmsge{u}.vx vd, va, x, v0.t
 | vmslt{u}.vx vd, va, x, v0.t +
--- a/src/asm-manual.adoc
+++ b/src/asm-manual.adoc
@@ -1091,8 +1091,6 @@ srli rd, rd, XLEN - 32
 |vmsgeu.vi vd, va, i, vm      | vmsgtu.vi vd, va, i-1, vm                                     | Vector >= Immediate, unsigned|
 |vmsge.vv vd, va, vb, vm      | vmsle.vv vd, vb, va, vm                                       | Vector >= Vector|
 |vmsgeu.vv vd, va, vb, vm     | vmsleu.vv vd, vb, va, vm                                      | Vector >= Vector, unsigned |
-|vmsge.vx vd, va, x, vm       | vmsle.vx vd, x, va, vm                                        | Vector >= scalar|
-|vmsgeu.vx vd, va, x, vm      | vmsleu.vx vd, x, va, vm                                       | Vector >= scalar, unsigned|
 |vmsgt.vv vd, va, vb, vm      | vmslt.vv vd, vb, va, vm                                       | Vector > Vector|
 |vmsgtu.vv vd, va, vb, vm     | vmsltu.vv vd, vb, va, vm                                      | Vector > Vector, unsigned|
 |vmslt.vi vd, va, i, vm       | vmsle.vi vd, va, i-1, vm                                      | Vector < immediate|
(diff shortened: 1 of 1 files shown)


cmuellner (changes requested):
This list is incomplete (e.g. `vl1r.v`/`vl2r.v`/`vl4r`/`vl8r.v` are missing).

To catch them all, either search for "Pseudoinstruction" in the specification or grep for `INSN_ALIAS` in Binutils.

AFOliveira:
@cmuellner  I added the ones you mentioned, but I couldn't really find many more.
I described them as "Equal to vl..." since any other description seemed like to long, but I can also change that if needed.

AFOliveira:
@cmuellner just a quick ping, is there anything else I should enhance in the PR?

AFOliveira:
👀 :)

src/asm-manual.adoc
```diff
@@ -1079,6 +1079,37 @@ srli rd, rd, XLEN - 32
...
+|vmfgt.vv vd, va, vb, vm      | vmflt.vv vd, vb, va, vm                                       | Vector Floating-point >|
+|vmmv.m vd, vs                | vmand.mm vd, vs, vs                                           | Vector copy mask register |
+|vmnot.m vd, vs               | vmnand.mm vd, vs, vs                                          | Vector invert mask bits|
+|vmset.m vd                   | vmxnor.mm vd, vd, vd                                          | Vector set all mask bits|
+|vmsge.vi vd, va, i, vm       | vmsgt.vi vd, va, i-1, vm                                      | Vector >= Immediate|
+|vmsgeu.vi vd, va, i, vm      | vmsgtu.vi vd, va, i-1, vm                                     | Vector >= Immediate, unsigned|
+|vmsge.vv vd, va, vb, vm      | vmsle.vv vd, vb, va, vm                                       | Vector >= Vector|
+|vmsgeu.vv vd, va, vb, vm     | vmsleu.vv vd, vb, va, vm                                      | Vector >= Vector, unsigned |
+|vmsge.vx vd, va, x, vm       | vmsle.vx vd, x, va, vm                                        | Vector >= scalar|
```
topperc, line 1094: This is the not the correct expansion for vmsge.vx

From the spec.

```
  The RISC-V Instruction Set Manual Volume I | © RISC-V International
pseudoinstruction: vmsge{u}.vx vd, va, x, v0.t, vt
    expansion: vmslt{u}.vx vt, va, x;  vmandn.mm vt, v0, vt;  vmandn.mm vd, vd,
  v0;  vmor.mm vd, vt, vd
    The vt argument to the pseudoinstruction must name a temporary vector
  register that is
not same as vd and which will be clobbered by the pseudoinstruction
```

AFOliveira: @topperc  How can I define this different implementations and follow manual style? Shall I just do a * like in li?

topperc: There are 3 different pseudoinstructions there that should each have their own row

vmsge{u}.vx vd, va, x
vmsge{u}.vx vd, va, x, v0.t // add a note in the last column that vd can't be v0
vmsge{u}.vx vd, va, x, v0.t, vt // you can write the 4 instruction expansion and a note that 2 instructions aren't needed if vd==v0.

AFOliveira: @topperc In the last commit I kept it exactly as described in the ISA Manual because I feel the 1:1 mapping is easier to understand, is this ok?

src/asm-manual.adoc
```diff
@@ -1079,6 +1079,64 @@ srli rd, rd, XLEN - 32
...
+|vnot.v vd,vs,vm              | vxor.vi vd, vs, -1, vm                                        | Vector not |
+|vncvt.x.x.w vd,vs,vm         | vnsrl.wx vd,vs,x0,vm                                          | Vector narrow convert element |
+|vwcvt.x.x.v vd,vs,vm         | vwadd.vx vd,vs,x0,vm                                          | Vector widen convert, integer-integer|
+|vwcvtu.x.x.v vd,vs,vm        | vwaddu.vx vd,vs,x0,vm                                         | Vector widen convert, integer-integer, unsigned|
+|vl1r.v v3, x0                | vl1re8.v v3, x0                                               | Equal to vl1re8.v |
+|vl2r.v v2,x0                 | vl2re8.v v2, x0                                               | Equal to vl2re8.v |
+|vl4r.v v4,x0                 | vl4re8.v v4, x0                                               | Equal to vl4re8.v |
+|vl8r.v v8,x0                 | vl8re8.v v8, x0                                               | Equal to vl8re8.v |
+
+|vmsge{u}.vx vd, va, x|
```
topperc, line 1108: This form isn't implemented in the assembler because it's impossible to know if `x>minimum` without doing more analysis than an assembler is capable of.

src/asm-manual.adoc
topperc: Yes

src/asm-manual.adoc
AFOliveira: Should be fixed now, thanks!

AFOliveira: My bad, I messed up when commiting, now should be fixed. Sorry for the mess up.

topperc (approved):
LGTM

cmuellner:
I checked the list against Binutils and found the following aliases in Binutils, which are not in the list:
* vmcpy.m (vmand.mm)
* vle1.v (vlm.v)
* vse1.v (vsm.v)

topperc:
> I checked the list against Binutils and found the following aliases in Binutils, which are not in the list:
> 
> * vmcpy.m (vmand.mm)
> * vle1.v (vlm.v)
> * vse1.v (vsm.v)

Those are old names instructions that were renamed. The aliases are for backwards compatibility. I'm not sure we should encourage their use.

cmuellner:
Ok, I wasn't aware of this.
Let's merge this then.

cmuellner (approved):

PR #114, opened 2024-11-23
Add CI workflow to create PDF/HTML

axel-h:
Address https://github.com/riscv-non-isa/riscv-asm-manual/issues/112

- Add CI workflow to create PDF/HTML and draft release for every push to main
- improve PDF content layout

See here how it looks like: https://github.com/axel-h/riscv-asm-manual/releases

### DIFF
--- a/.github/workflows/build-doc.yml
+++ b/.github/workflows/build-doc.yml
@@ -1,0 +1,71 @@
+name: Create Document
+
+on:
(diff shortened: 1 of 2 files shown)


wmat (approved):
I don't see any reason not to merge this.  All risc-v document repos should have the same CI.

PR #120, opened 2024-12-19
Eliminating trailing whitespaces

cmuellner:
A recent PR introduced a few trailing whitespaces
and the CI complains about it.
Let's fix that.


cmuellner:
@apazos, please review.

PR #124, opened 2025-02-17
Add fgt.[hsd] and fge.[hsd] pseudo-instructions

kito-cheng:
Those two pseudo-instructions are trivial and also already implemented in the
binutils[1] and LLVM[2].

[1] https://github.com/bminor/binutils-gdb/commit/e23eba971dd409b999dd83d8df0f842680c1c642
[2] https://github.com/llvm/llvm-project/commit/84080793c56558b6e677fa70577e87e2e8404816

NOTE: bintuils supports those 2 kind of pseudo-instructions since RISC-V support began.

---

Also come with 2 small patch for tweak the document:
- `Add .h variant of floating point pesudo instructions`
- `Use fr[ds] for FPR to prevent confuse with GPR for floating point pesudo instructions`


kito-cheng:
cc @asb

aswaterman (approved):

cmuellner (approved):
LGTM

PR #128, opened 2025-04-09
[Zilsd] Clarify Symbol Pseudos for Zilsd

lenary:
The existing documentation didn't mention that `ld` and `sd` symbol pseudos aren't available on RV32. This change clarifies that these are available, but denote pair load/stores using the Zilsd instructions.

### DIFF
--- a/src/asm-manual.adoc
+++ b/src/asm-manual.adoc
@@ -784,6 +784,10 @@ as seen by `objdump`:
       1c: R_RISCV_PCREL_LO12_S  .L0
 ----
 
+On RV32, `ld <rd>, <symbol>` and `sd <rd>, <symbol>, <rt>` can be used to
+generate a sequence including the `ld` and `sd` instructions in Zilsd -- in
+these cases, `<rd>` denotes the even GPR in an even-odd GPR pair.
+
 == Constants
 
 The following example shows loading a constant using the `%hi` and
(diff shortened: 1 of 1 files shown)


lenary:
These have been proposed for:
- LLVM: https://github.com/llvm/llvm-project/pull/134950
- GCC: https://sourceware.org/pipermail/binutils/2025-March/140283.html

topperc (approved):
LGTM

cmuellner (approved):
LGTM

src/asm-manual.adoc
```diff
@@ -784,6 +784,10 @@ as seen by `objdump`:
       1c: R_RISCV_PCREL_LO12_S  .L0
 ----
 
+On RV32, `ld <rd>, <symbol>` and `sd <rd>, <symbol>, <rt>` can be used to
```
MaskRay, line 787: Is <symbol>+<offset> allowed?

(Allowed in gas and llvm but the llvm AsmParser is a bit of a hack.)

lenary: LLVM looks like it is passing the offset through to the relocation correctly, so I'm happy to allow this. I can clarify the wording to ensure it is, because there are other places that allow `<symbol>{+<offset>}` when they only say `<symbol>` right now.

lenary: I updated this to be `<symbol expression>` which can be symbol, offset, or symbol with offset.

MaskRay (approved):

cmuellner (approved):
We discussed this briefly in today's SIG Toolchain meeting and agreed there to land this.

PR #199, opened 2026-02-27
Add more jalr pseudoinstructions

topperc:
Fixes #198


lenary (approved):
These do indeed exist.

It would be good to more clearly work out which of these aliases should be canonical (printed), vs accepted (parsed).

lenary:
> But I can say for these that they are canonical. There are more non-canonical jalr pseudoinstructions I did not include.

I guess I would expect the canonical syntax to have been closer to `jalr <optional-reg + comma> <optional-offset>(reg)` but i guess the fact that the second register is a memory address is overlooked here. I realise `jalr` is common enough that changing this is a bit pointless at this point.

cmuellner (approved):
LGTM

PR #216, opened 2026-04-02
docs: add missing call rd pseudoinstruction table entry

### LINKED ISSUE #134: call <rd>, <symbol> is missing from pseudo instructions table
The call pseudo instruction is listed as follows:

> call offset | auipc x1, offset[31:12]jalr x1, x1, offset[11:0] | Call far-away subroutine

Earlier, two flavors of call are defined

> call <symbol>: call away subroutine
> 
> call <rd>, <symbol>: call away subroutine

call rd, offset should be added.

I was also if it is mandatory to assemble this into the given sequence, or if it is possible to use a jump and link with immediate if the subroutine is less far away as expected.

xinx1ang:
## Summary
- add the missing `call rd, offset` entry to the standard pseudoinstruction table
- keep the table aligned with the earlier Function Calls section and issue #134

## Why
The manual already documents both `call <symbol>` and `call <rd>, <symbol>` in the Function Calls section, but the summary pseudoinstruction table only listed the plain `call offset` form. Adding the `call rd, offset` row makes the table consistent and addresses #134.

Closes #134

### DIFF
--- a/src/asm-manual.adoc
+++ b/src/asm-manual.adoc
@@ -1168,6 +1168,18 @@ jalr x0, offset[11:0](rt)
 |Jump to far-away label
 |
 
+|call offset
+|auipc x1, offset[31:12] +
+jalr x1, x1, offset[11:0]
+|Call far-away subroutine
+|
+
+|call rd, offset
+|auipc rd, offset[31:12] +
+jalr rd, rd, offset[11:0]
+|Call far-away subroutine
+|Uses `rd` to save the return address.
+
 |jal offset                   | jal x1, offset                                                | Jump and link |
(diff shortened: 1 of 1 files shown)


topperc (approved):
LGTM

cmuellner:
Thanks for looking at this.

#134 has been addressed by #139.

I think the changes added in this PR are already on lines 1233 and 1239 below.
The only difference that I see is that the expanded instruction sequence uses "rd, rd, offset" (in PR) instead of "rd, offset(rd)" (in repo).

apazos:
I agree the pseudo instructions are already in the document.

cmuellner:
I'm closing this as we already have this pseudos.

PR #219, opened 2026-05-07
Clarify Store Pseudo Temporary Register

### LINKED ISSUE #218: Clarifying the Scratch/Temporary/Clobber Register in the Store symbolic Pseudos
The store pseudos that take a symbol also have to take a "scratch register" for computing the address to store in.

```
  sw rs, symbol, rt
  # expands to =>
  .Lpcrel:
  auipc rt, %pcrel_hi(symbol)
  sw rs, %pcrel_lo(Lpcrel)(rt)
```

One thing that isn't clear is the liveness of `rt`, the scratch register.

I think the reality is that `rt` is no longer live, because it is a semi-initialised pc-relative value, so you have to know quite a lot about what is being accessed (i.e. the alignment of the symbol) to reuse that value in another store. IMO, sequences like this would be written more clearly with one auipc and multiple register-based stores, but that opinion may not be shared.

I think we don't want users to reuse `rt`, and we should probably document this, but I welcome feedback before I turn this into a PR.

lenary:
Closes #218

### DIFF
--- a/src/asm-manual.adoc
+++ b/src/asm-manual.adoc
@@ -799,7 +799,7 @@ footnote:fn-1[the first operand is implicitly used as a scratch register.]
 * `l{bu|hu|wu} <rd>, <symbol expression>`: load unsigned byte, half word, or word from global{empty}
 footnote:fn-1[]
 * `s{b|h|w|d} <rd>, <symbol expression>, <rt>`: store byte, half word, word or double word to global{empty}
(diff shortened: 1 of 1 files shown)


topperc (approved):
LGTM

kito-cheng (approved):
LGTM :)

cmuellner (approved):
LGTM

Issue #78, opened 2022-04-23
wrong instruction generation


jnk0le:
in "GOT-indirect addressing"

>1:
	auipc	a0, %got_pcrel_hi(msg + 1)
	ld	a0, %pcrel_lo(1b)(a0)

>Which generates the following assembler output and relocations as seen by objdump:

>0000000000000000 <.text>:
   0:	00000517          	auipc	a0,0x0
			0: R_RISCV_GOT_HI20	msg+0x1
   4:	00050513          	mv	a0,a0
			4: R_RISCV_PCREL_LO12_I	.L1

`ld` should not expand into `mv`, looks like copypasted from section above

pdfrod:
This doesn't seem fixed? I'm still seeing `mv  a0,a0` in [GOT-indirect addressing
](https://github.com/riscv-non-isa/riscv-asm-manual/blob/3b06fa93fd52f277e8e0dc180bbb2ceec1ff5957/src/asm-manual.adoc#got-indirect-addressing) section.

aswaterman:
Agreed, that line should be replaced with `   4: 00053503            ld a0,0(a0)`.  The rest is correct.

jnk0le:
one thing is that this could be actual intermediate output from compilers which gets "relaxed" at later stages.

aswaterman:
Yeah, but not in this particular situation, AFAICS.

PR #222, opened 2026-06-14
Fix GOT-indirect addressing example output

pz9115:
The GOT-indirect addressing example uses:

    ld a0, %pcrel_lo(1b)(a0)

so the objdump output should show an RV64 load from the GOT entry rather than
`mv a0,a0`, which appears to have been copied from the previous PC-relative
addressing example.

Also clarify that the shown output is for RV64; RV32 uses `lw` instead.

Verified with riscv64-unknown-elf-as and riscv64-unknown-elf-objdump.

A Compiler Explorer reproduction is also available:
https://godbolt.org/z/Tsse84qae

Ref #78.

### DIFF
--- a/src/asm-manual.adoc
+++ b/src/asm-manual.adoc
@@ -582,15 +582,15 @@ The following example shows how to load an address from the GOT:
   ld  a0, %pcrel_lo(1b)(a0)
 ----
 
-Which generates the following assembler output and relocations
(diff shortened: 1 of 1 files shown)


topperc (approved):
LGTM

cmuellner (approved):
LGTM

PR #227, opened 2026-08-27
Add optional register argument to 'tail' pseudo.

topperc:
The lpad instructions from Zicfilp are encoded using an auipc hint encoding. The intent is to be able to instrument a binary and have it run correctly on a CPU that doesn't support Zicfilp.

In order to do that, compilers need to generate instrumented code without Zicfilp appearing in the ISA string. This means we need to be able to emit a software-guarded tail call using x7 as the register independent of the ISA string.

I didn't remove the Zicfilp behavior for the default register since it is already implemented in the toolchains.

### DIFF
--- a/src/asm-manual.adoc
+++ b/src/asm-manual.adoc
@@ -908,8 +908,10 @@ footnote:fn-3[`ra` is implicitly used to save the return address.]
 footnote:fn-4[similar to `call <symbol>`, but `<rd>` is used to save the return address instead.]
 * `tail <symbol>`: tail call away subroutine{empty}
 footnote:fn-5[If the `Zicfilp` extension is available, `t2` is implicitly used as a scratch register. Otherwise,`t1` is implicitly used as a scratch register.]
+* `tail <symbol>, <rt>`: tail call away subroutine{empty}
+footnote:fn-6[`<rt>` is implicitly used as a scratch register.]
 * `jump <symbol>, <rt>`: jump to far-away label{empty}
-footnote:fn-6[similar to `tail <symbol>`, but `<rt>` is used as the scratch register instead.]
+footnote:fn-7[similar to `tail <symbol>`, but `<rt>` is used as the scratch register instead.]
 
 The following example shows how these pseudoinstructions are used:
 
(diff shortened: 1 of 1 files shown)


topperc:
CC: @kito-cheng

topperc:
CC: @aswaterman do you have any opinions on naming here?

aswaterman:
What about just using `jump foo, x7` instead of defining something new?

Or, failing that, `tail.x7`?

topperc:
> What about just using `jump foo, x7` instead of defining something new?
> 
> Or, failing that, `tail.x7`?

I've modified the proposal to add an optional register to the existing tail pseudo.

aswaterman:
Ah yeah, that makes sense.

lenary:
I don't think tail.x7 is good because it doesn't clarify why x7. Tail with an extra register is like call with an extra register so ok on that front (as long as those two align). Jump is not incorrect but tail seems a better fit for clarifying the intention.

topperc:
> I don't think tail.x7 is good because it doesn't clarify why x7. Tail with an extra register is like call with an extra register so ok on that front (as long as those two align). Jump is not incorrect but tail seems a better fit for clarifying the intention.

call has the register first, I put the register second on this. Given the optional nature that seemed friendlier to the parser.

aswaterman:
For call, the role is that of a destination register, whereas here, it is that of a temporary. By existing convention, it should go at the end for tail.

lenary:
Ok, that feels justified enough. 

Let's one day write some of these conventions down :)

aswaterman:
BTW, was there some reason not to just change the register to x7 unconditionally?  The explicit scheme is probably better, especially since we'll get an error if the assembler doesn't support it, rather than silently using the wrong register. I'm just curious.

topperc:
> BTW, was there some reason not to just change the register to x7 unconditionally? The explicit scheme is probably better, especially since we'll get an error if the assembler doesn't support it, rather than silently using the wrong register. I'm just curious.

It would create a version lock between clang and binutils if `-fno-integrated-as` is used. If the wrong version of binutils is used, you wouldn't know until runtime on a Zicfilp enabled system.

Paranoia about changing the binary generated for an instruction mnemonic that has been around for a long time.

aswaterman:
Ack.

topperc:
Ping @lenary @cmuellner @lenary

lenary (approved):
LGTM

kito-cheng (approved):
LGTM, and also binutils patch: https://patchwork.sourceware.org/project/binutils/patch/20260914095359.832847-1-kito.cheng@sifive.com/

